PR #37, opened 2020-08-30
Reduce "clean build" time by 3 times

ef1rspb:
Use cocopods-binary plugin to generate .framework for desired pods


ef1rspb:
Motivation: switching between branches often requires xcode to recompile pods, so the developer is going to sit back. [Cocoapods binary](https://github.com/leavez/cocoapods-binary) prebuild desired pods into frameworks. Clean build reduced from 140 seconds to 47 on my macbook pro.

ERussel (commented):
Good point! Thank you! But I think we need to update CI to be able to merge this improvement.

ERussel:
Closing now until we will have time to improve CI/CD

PR #52, opened 2020-10-02
fix cell xib linking

ERussel:


antonkhvorov (approved):

antonkhvorov:
Its always xibs what causing the problems

PR #60, opened 2020-10-14
FLW-389 Existentional deposit check

ERussel:
- add view to display balance details on transfer screen
- add check for existential deposit during transfer


fearless/Modules/Wallet/Commands/TransferConfirmCommand.swift
```diff
@@ -0,0 +1,69 @@
+import Foundation
+import CommonWallet
+import SoraFoundation
+
+final class TransferConfirmCommand: WalletCommandDecoratorProtocol {
+    var undelyingCommand: WalletCommandProtocol?
```
ArsenyZ, line 6: underlying может быть?

ArsenyZ (approved):

fearless/Modules/Wallet/Commands/TransferConfirmCommand.swift
ERussel: Да, спасибо. Поправлю.

PR #98, opened 2020-12-16
Ramp integration

ERussel:
- buy logic implementation add


fearless/Common/PurchaseProvider/RampProvider.swift
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    func buildPurchaseAction(for chain: Chain,
+                             assetId: WalletAssetId?,
+                             address: String) -> [PurchaseAction] {
+        let optionUrl: URL?
+
+        switch chain {
+        case .polkadot:
+            optionUrl = buildURLForToken("DOT", address: address)
```
antonkhvorov, line 27: nice

fearless/Common/PurchaseProvider/RampProvider.swift
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    private func buildURLForToken(_ token: String, address: String) -> URL? {
+        var components = URLComponents(string: Self.baseUrlString)
+
+        var queryItems = [
+            URLQueryItem(name: "swapAsset", value: token),
+            URLQueryItem(name: "userAddress", value: address),
+            URLQueryItem(name: "hostApiKey", value: Self.pubToken)
```
antonkhvorov, line 48: in the future, if we will want to support Staging envs for payment providers, this probably should be changed

antonkhvorov (approved):

fearless/Common/PurchaseProvider/RampProvider.swift
ERussel: Actually not, as we can see Self.pubToken and Self.baseUrl are static vars and can be wrapped with preprocessing macros to have different values for different environments

PR #103, opened 2020-12-18
v1.1 Buy DOT tokens feature

ERussel:
- add integration with Ramp
- buy DOT feature add
- add error when amount which is less than existential deposit is sent to fresh account


antonkhvorov (approved):
v1.1 lets go 🎉

PR #112, opened 2021-02-01
FLW-328 iOS. On the "Add account" button shows the system element

stas-litvinov:
Replace image with arrow with a new one in AccountManagementViewController.xib


ERussel (approved):

stas-litvinov:
CI did not work with PR from forked project

PR #118, opened 2021-02-11
FLW-539 iOS: Warning about Phishing address

stas-litvinov:
fetch address.json from https://polkadot.js.org/phishing/address.json
parse address.json and store public keys in local db
on Send flow, on Recipient screen after address is entered and user wants to proceed with Sending tokens on that address, raise warning message
rise this warning message again on Send screen before Confirmation screen

update localisations


fearless/Common/Network/GitHubOperationFactory.swift
```diff
@@ -0,0 +1,52 @@
...
+        let resultFactory = AnyNetworkResultFactory<[PhishingItem]> { data in
+            var phishingItems: [PhishingItem] = []
+
+            if let json = try JSONSerialization.jsonObject(with: data,
+                                                           options: [.mutableContainers]) as? [String: AnyObject] {
+                for (key, value) in json {
+                    if let addresses = value as? [String] {
+                        for address in addresses {
+                            do {
+                                let typeRawValue = try SS58AddressFactory().type(fromAddress: address)
```
ERussel, line 28: I guess we need to create the factory once before the loop

fearless/Common/Network/GitHubOperationFactory.swift
```diff
@@ -0,0 +1,52 @@
...
+            request.httpMethod = HttpMethod.get.rawValue
+            return request
+        }
+
+        let resultFactory = AnyNetworkResultFactory<[PhishingItem]> { data in
+            var phishingItems: [PhishingItem] = []
+
+            if let json = try JSONSerialization.jsonObject(with: data,
+                                                           options: [.mutableContainers]) as? [String: AnyObject] {
+                for (key, value) in json {
```
ERussel, line 24: It is not so critical but why not to use functional style compactMap/flatMap/map here. The main benefit is that we are sure that we put an item at most once to the list in the loop.

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
...
+import RobinHood
+
+class GitHubPhishingAPIService: ApplicationServiceProtocol {
+    lazy var endPoint: String = { return "https://polkadot.js.org/phishing/address.json" }()
+
+    private var operation: BaseOperation<[PhishingItem]>
+    private var logger: Logger
+    private var storage: CoreDataRepository<PhishingItem, CDPhishingItem>
+
+    init(operation: BaseOperation<[PhishingItem]>,
```
ERussel, line 11: What if we will need to execute an opreration multiple time. I guess we need to pass an operation factory here

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
...
+    private func clearConnection() {
+        operation.cancel()
+    }
+
+    private func setupConnection() {
+        operation.completionBlock = { [self] in
+            do {
+                if let phishingItems = try operation.extractResultData() {
+                    let deleteOperation = storage.deleteAllOperation()
+                    OperationManagerFacade.sharedManager.enqueue(operations: [deleteOperation], in: .sync)
```
ERussel, line 57: I think we need to pass OperationManager in the constructor of the service

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private func clearConnection() {
+        operation.cancel()
+    }
+
+    private func setupConnection() {
+        operation.completionBlock = { [self] in
+            do {
+                if let phishingItems = try operation.extractResultData() {
+                    let deleteOperation = storage.deleteAllOperation()
```
ERussel, line 56: I think we should use replace operation instead of delete+save

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
...
+
+        clearConnection()
+    }
+
+    private func clearConnection() {
+        operation.cancel()
+    }
+
+    private func setupConnection() {
+        operation.completionBlock = { [self] in
```
ERussel, line 53: I think we can use dependencies to chain operations.

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+import RobinHood
+
+class GitHubPhishingAPIService: ApplicationServiceProtocol {
+    lazy var endPoint: String = { return "https://polkadot.js.org/phishing/address.json" }()
+
+    private var operation: BaseOperation<[PhishingItem]>
+    private var logger: Logger
```
ERussel, line 8: I think we need to pass LoggerProtocol

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+import RobinHood
+
+class GitHubPhishingAPIService: ApplicationServiceProtocol {
+    lazy var endPoint: String = { return "https://polkadot.js.org/phishing/address.json" }()
```
ERussel, line 5: I suggest to move this to ApplicationConfig

fearless/Modules/Wallet/Commands/TransferConfirmCommandProxy.swift
```diff
@@ -0,0 +1,75 @@
...
+        self.calleeCommand = TransferConfirmCommand(payload: payload,
+                                                  localizationManager: localizationManager,
+                                                  commandFactory: commandFactory)
+    }
+
+    func execute() throws {
+        let destinationKey = calleeCommand.payload.transferInfo.destination
+
+        let storage: CoreDataRepository<PhishingItem, CDPhishingItem> =
+            SubstrateDataStorageFacade.shared.createRepository()
```
ERussel, line 28: I suggest to pass repository in contstuctor

fearless/Modules/Wallet/Commands/TransferConfirmCommandProxy.swift
```diff
@@ -0,0 +1,75 @@
...
+            DispatchQueue.main.async {
+                self.handleAccountFetch(result: fetchOperation.result)
+            }
+        }
+
+        OperationManagerFacade.sharedManager.enqueue(operations: [fetchOperation], in: .sync)
+    }
+
+    private func handleAccountFetch(result: Result<PhishingItem?, Error>?) {
+        switch result {
```
ERussel, line 42: I think the better way is to use ```extractResultData(...)``` method from operation

fearless/Modules/Wallet/Commands/TransferConfirmCommandProxy.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    var undelyingCommand: WalletCommandProtocol? {
+        get { calleeCommand.undelyingCommand }
+        set { calleeCommand.undelyingCommand = newValue }
+        }
+
+    let logger = Logger.shared
+
+    init(payload: ConfirmationPayload,
+         localizationManager: LocalizationManagerProtocol,
```
ERussel, line 17: I think we need to get rid of localization manager is pass local here.

fearless/Modules/Wallet/Contacts/ContactViewModel.swift
```diff
@@ -18,20 +21,71 @@ final class ContactViewModel: ContactsLocalSearchResultProtocol {
...
+
+        case .failure(let error):
+            logger.error(error.localizedDescription)
+
+        case .none:
+            logger.error("Scam account fetch operation cancelled")
+        }
     }
+
+    private func showWarning() {
```
ERussel, line 74: I think I saw it before :)

ERussel (changes requested):
I suggest to make PhishingCheckExecutor class that encapsulates logic for checking the address and presenting the alert. Then we can reuse it both in ContactCommandDelegate and in TransferConfirmationProxy.

ERussel (approved):

PR #122, opened 2021-02-16
FLW-539 iOS: Warning about Phishing address

stas-litvinov:
refactor issues left after PR #118:

- replace nested loops with flatMap/map
- add contacts view model delegate proxy
- use use replace operation instead of delete+save
- pass OperationManager in the constructor of GitHubPhishingAPIService
- move repository declaration outside TransferConfirmCommandProxy class
- put operationManager to GitHubPhishingAPIService initializer
- put SS58AddressFactory outside the loop
- change Logger to LoggerProtocol
- put phishing list address into AppConfig


fearless/Common/Network/GitHubOperationFactory.swift
```diff
@@ -24,34 +24,45 @@ class GitHubOperationFactory: GitHubOperationFactoryProtocol {
...
-                                                    publicKey: accountId.toHex())
-                            phishingItems.append(item)
+
+            let phishingItems = json.flatMap { (key, value) -> [PhishingItem] in
+                if let publicKeys = value as? [String] {
+                    let items = publicKeys.map {
+                        if let publicKey = self.getPublicKey(from: $0, using: addressFactory) {
+                            return publicKey
                         }
+                        return ""
```
ERussel, line 35: I think we can use compactMap instead of map and reduce the closure to just getPublicKey(...)

fearless/Common/Services/GitHubPhishingService/PhishingCheckExecutor.swift
```diff
@@ -0,0 +1,60 @@
...
+import CoreData
+import SoraFoundation
+import CommonWallet
+
+protocol PhishingCheckExecutorProtocol {
+    func checkPhishing(publicKey: String, walletAddress displayName: String)
+}
+
+class PhishingCheckExecutor<T: Identifiable, U: NSManagedObject>: PhishingCheckExecutorProtocol {
+    private var storage: CoreDataRepository<T, U>
```
ERussel, line 12: It is more convenient to use AnyDataProviderRepository. In this case one doesn't care about database type

fearless/Modules/Wallet/Contacts/ContactViewModel.swift
```diff
@@ -37,55 +37,7 @@ extension ContactViewModel: WalletCommandProtocol {
     var command: WalletCommandProtocol? { self }
 
     func execute() throws {
```
ERussel, line 39: As we discussed it is better to pass a command instead of delegate proxy. So you don't need to store delegate proxy as a field in the factory

stas-litvinov: The problem is that command is a get-only property and it returns viewModel. I don't quite understand how to change it to make it do phishing check in one case and to keep its behaviour in all the other cases. Let's discuss

fearless/Common/Services/GitHubPhishingService/GitHubPhishingAPIService.swift
```diff
@@ -44,26 +51,24 @@ class GitHubPhishingAPIService: ApplicationServiceProtocol {
...
+            return phishingItem
         }
 
-        OperationManagerFacade.sharedManager.enqueue(operations: [operation], in: .sync)
+        replaceOperation.addDependency(networkOperation)
+
+        let operationWrapper = CompoundOperationWrapper(targetOperation: replaceOperation,
+                                        dependencies: [networkOperation])
+
+        OperationManagerFacade.sharedManager.enqueue(operations: operationWrapper.allOperations, in: .sync)
```
ERussel, line 72: We have passed operation manager but don't use it

stas-litvinov: Opps. Fixed

fearless/Modules/Wallet/Contacts/ContactsViewModelFactory.swift
```diff
@@ -22,13 +24,36 @@ final class ContactsViewModelFactory: ContactsFactoryWrapperProtocol {
...
 
-            return ContactViewModel(firstName: contact.firstName,
-                                    lastName: contact.lastName,
-                                    accountId: contact.accountId,
-                                    image: icon,
-                                    name: contact.firstName,
-                                    delegate: delegate,
-                                    commandFactory: commandFactory)
+            let storage: CoreDataRepository<PhishingItem, CDPhishingItem> =
+                SubstrateDataStorageFacade.shared.createRepository()
```
ERussel, line 28: I suggest to pass facade as a parameter to the factory

fearless/Modules/Wallet/Contacts/ContactsViewModelFactory.swift
```diff
@@ -22,13 +24,36 @@ final class ContactsViewModelFactory: ContactsFactoryWrapperProtocol {
...
+
+            let viewModel = ContactViewModel(
+                firstName: contact.firstName,
+                lastName: contact.lastName,
+                accountId: contact.accountId,
+                image: icon,
+                name: contact.firstName)
+
+            let nextAction = {
+                delegate?.didSelect(contact: viewModel)
```
ERussel, line 38: I suggest to capture delegate as weak

fearless/Modules/Wallet/WalletCommandDecorator.swift
```diff
@@ -18,8 +19,13 @@ final class WalletCommandDecoratorFactory: WalletCommandDecoratorFactoryProtocol
     func createTransferConfirmationDecorator(with commandFactory: WalletCommandFactoryProtocol,
                                              payload: ConfirmationPayload)
         -> WalletCommandDecoratorProtocol? {
-        TransferConfirmCommandProxy(payload: payload,
+
+        let storage: CoreDataRepository<PhishingItem, CDPhishingItem> =
+            SubstrateDataStorageFacade.shared.createRepository()
```
ERussel, line 24: I suggest to pass facade as parameter

ERussel (approved):

PR #124, opened 2021-02-23
Fast review: fix typos, omit "public"

ef1rspb:


fearless/Common/Scale/Types/AccountInfo.swift
```diff
@@ -49,7 +49,7 @@ struct AccountData: ScaleDecodable {
     }
 }
 
-struct Balance: ScaleCodable {
```
ERussel, line 52: Let's leave ScaleCodable as we have manual implementation for encoding.

fearless/Common/Network/Subscan/Data/PriceData.swift
```diff
@@ -1,13 +1,13 @@
 import Foundation
 
-struct PriceData: Codable, Equatable {
```
ERussel, line 3: Now we a caching this entity. So we need to implement both Codable and Equatable

ERussel (changes requested):
Finally, we have fixed our ci to support external PRs. But we need to fix synchronise with develop to merge the branch.

ERussel (approved):

PR #127, opened 2021-02-26
FLW-579 iOS: Elected validators in Era service

ERussel:
- era validation service implementation add: main purpose is to fetch validators list when era changes and provide it to other services on request
- refactor service loading by introducing service coordinator and moving initialization of the services to the correponding factories
- add operations to create storage keys based on runtime: MapKeyEncodingOperation, DoubleMapKeyEncodingOperation
- add storage response decoding operation based on runtime
- add storage request factory to incapsulate pipeline: prepare key, get data, decode data

### DIFF
--- a/fearless/Common/Operation/StorageDecodingOperation.swift
+++ b/fearless/Common/Operation/StorageDecodingOperation.swift
@@ -1,0 +1,97 @@
+import Foundation
+import FearlessUtils
+import RobinHood
+
+enum StorageDecodingOperationError: Error {
+    case missingRequiredParams
+    case invalidStoragePath
+}
+
+final class StorageDecodingOperation<T: Decodable>: BaseOperation<T> {
+    var data: Data?
+    var codingFactory: RuntimeCoderFactoryProtocol?
+
+    let path: StorageCodingPath
+
+    init(path: StorageCodingPath, data: Data? = nil) {
+        self.path = path
+        self.data = data
+
+        super.init()
+    }
+
+    override func main() {
+        super.main()
+
+        if isCancelled {
+            return
+        }
+
+        if result != nil {
+            return
+        }
+
+        do {
+            guard let data = data, let factory = codingFactory else {
+                throw StorageDecodingOperationError.missingRequiredParams
+            }
+
+            guard let entry = factory.metadata.getStorageMetadata(in: path.moduleName,
+                                                                  storageName: path.itemName) else {
+                throw StorageDecodingOperationError.invalidStoragePath
+            }
+
+            let decoder = try factory.createDecoder(from: data)
+            let item: T = try decoder.read(of: entry.type.typeName)
+            result = .success(item)
+        } catch {
+            result = .failure(error)
+        }
+    }
+}
+
+final class StorageDecodingListOperation<T: Decodable>: BaseOperation<[T]> {
+    var dataList: [Data]?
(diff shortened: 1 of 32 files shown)


fearless/Common/Operation/StorageDecodingOperation.swift
```diff
@@ -0,0 +1,97 @@
...
+import FearlessUtils
+import RobinHood
+
+enum StorageDecodingOperationError: Error {
+    case missingRequiredParams
+    case invalidStoragePath
+}
+
+final class StorageDecodingOperation<T: Decodable>: BaseOperation<T> {
+    var data: Data?
```
ef1rspb, line 11: let?

ERussel: This dependency can come from another operation, so we need an opportunity to update this in config closure

fearless/Common/Operation/StorageRequestFactory.swift
```diff
@@ -0,0 +1,181 @@
...
+            } catch {
+                decodingOperation.result = .failure(error)
+            }
+        }
+
+        decodingOperation.addDependency(queryOperation)
+
+        let mapOperation = ClosureOperation<[StorageResponse<T>]> {
+            let result = try queryOperation.extractNoCancellableResultData()
+            let keysData = result
```
ef1rspb, line 74: result.flatMap({ StorageUpdateData(update: $0).changes }) is computed 3 times. Suggest define 
```suggestion
            let resultChangesData = result.flatMap { StorageUpdateData(update: $0).changes }
```

fearless/Common/Operation/StorageRequestFactory.swift
```diff
@@ -0,0 +1,181 @@
...
+                                        dependencies: dependencies)
+    }
+
+    func queryItems<K, T>(engine: JSONRPCEngine,
+                          keyParams: @escaping  () throws -> [K],
+                          factory: @escaping  () throws -> RuntimeCoderFactoryProtocol,
+                          storagePath: StorageCodingPath)
+    -> CompoundOperationWrapper<[StorageResponse<T>]> where K: Encodable, T: Decodable {
+
+        let currentRemoteFactory = remoteFactory
```
ef1rspb, line 117: currentRemoteFactory is used just once

fearless/Common/Operation/StorageRequestFactory.swift
ERussel: Fixed

fearless/Common/Operation/StorageRequestFactory.swift
ERussel: Fixed

fearless/Common/Services/ServiceCoordinator.swift
```diff
@@ -0,0 +1,99 @@
...
+import SoraKeystore
+import SoraFoundation
+
+protocol ServiceCoordinatorProtocol: ApplicationServiceProtocol {
+    func updateOnAccountChange()
+    func updateOnNetworkChange()
+}
+
+final class ServiceCoordinator {
+    let webSocketService: WebSocketServiceProtocol
```
ArsenyZ, line 11: yeeeaah

ArsenyZ (approved):

PR #129, opened 2021-03-01
FLW-582 Calculation functions

stas-litvinov:
add staking reward calculator engine
add reward calculator retrieval service


fearless/Common/Services/ServiceCoordinator.swift
```diff
@@ -47,39 +52,88 @@ final class ServiceCoordinator {
...
             validatorService.update(to: chain, engine: engine)
             validatorService.setup()
         }
 
         gitHubPhishingAPIService.setup()
+
+        rewardCalculatorService.update(to: chain)
+        rewardCalculatorService.setup()
+
+        let validatorsOperation = validatorService.fetchInfoOperation()
```
ERussel, line 93: As we discussed we need to move this part to integration test

fearless/Common/Services/RewardCalculatorService/RewardCalculatorService.swift
```diff
@@ -0,0 +1,269 @@
...
+            pendingRequests.append(request)
+        }
+    }
+
+    private func deliver(snapshot: BigUInt, to request: PendingRequest) {
+        let eraOperation = eraValidatorsService.fetchInfoOperation()
+
+        eraOperation.completionBlock = {
+            dispatchInQueueWhenPossible(request.queue) {
+                if let result = try? eraOperation.extractResultData() {
```
ERussel, line 66: I suggest to wrap with switch..case and log the error

fearless/Common/Services/RewardCalculatorService/RewardCalculatorService.swift
```diff
@@ -0,0 +1,269 @@
...
+            logger.warning("Missing chain to proccess total issuance")
+            return
+        }
+
+        guard let totalIssuanceItem = totalIssuanceItem else {
+            return
+        }
+
+        let codingFactoryOperation = runtimeCodingService.fetchCoderFactoryOperation()
+        let decodingOperation = StorageDecodingOperation<String>(path: .totalIssuance,
```
ERussel, line 128: Now we can use StringScaleMapper (which is in develop already) to avoid manual mapping to the integer from string

ERussel (approved):

PR #135, opened 2021-03-08
Sprint 5 tech debt

ERussel:
- fix percentage bug on the main screen
- fix fee retrieval bug
- pass amount from main screen to setup amount screen
- fix test for main screen
- add integration test to benchmark staking setup after app launch


fearless/Modules/Staking/StakingAmount/StakingAmountProtocols.swift
```diff
@@ -57,5 +57,5 @@ protocol StakingAmountWireframeProtocol: AlertPresentable, ErrorPresentable,
 }
 
 protocol StakingAmountViewFactoryProtocol: class {
-	static func createView() -> StakingAmountViewProtocol?
+    static func createView(with amount: Decimal?) -> StakingAmountViewProtocol?
```
stas-litvinov, line 60: It is good to pass amount value but it is not recalculated when fee information is received. If amount is 100% of available balance then user will see insufficient funds alert which is confusing. I'd suggest checking the amount against (available balance - fee) whenever new fee is available
image
image

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -40,29 +40,37 @@ final class StakingMainPresenter {
...
+                                       targetReturn: monthlyReturn,
+                                       priceData: priceData)
 
-        let monthlyViewModel = viewModelFactory.createMonthlyRewardViewModel(amount: curAmount,
-                                                                             reward: reward,
-                                                                             priceData: priceData)
+            let yearlyViewModel = viewModelFactory
+                .createRewardViewModel(reward: (amount ?? 0.0) * yearlyReturn,
+                                       targetReturn: yearlyReturn,
+                                       priceData: priceData)
```
stas-litvinov, line 66: Looks like calculations are still incorrect: 
0.0002 KSM * 12 = 0.0024 KSM (expected 0.0036 KSM)
1,1% * 12 = 13,2% (expected 14,31%)
$0.06 * 12 = $0.72 (expected $0.83)
image

fearless/Common/Services/EraValidatorsService/EraValidatorService.swift
```diff
@@ -33,13 +33,13 @@ final class EraValidatorService {
...
     let operationManager: OperationManagerProtocol
-    let logger: LoggerProtocol
+    let logger: LoggerProtocol?
 
     init(storageFacade: StorageFacadeProtocol,
          runtimeCodingService: RuntimeCodingServiceProtocol,
          providerFactory: SubstrateDataProviderFactoryProtocol,
          operationManager: OperationManagerProtocol,
-         logger: LoggerProtocol) {
+         logger: LoggerProtocol? = nil) {
```
stas-litvinov, line 42: What is the purpose of making logger Optional?

fearless/Modules/Staking/StakingAmount/StakingAmountProtocols.swift
ERussel: For such a use case we need fee information on the main screen. Otherwise it will be strange that we display one amount on the main screen and another amount later.

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
ERussel: Did you check compound case?

fearless/Common/Services/EraValidatorsService/EraValidatorService.swift
ERussel: To allow not passing it in tests

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
stas-litvinov: Oh, I see now. In this case everything seems right. However, it might be confusing for a user. Let's discuss later

fearless/Common/Services/EraValidatorsService/EraValidatorService.swift
stas-litvinov: Thanks

fearless/Modules/Staking/StakingAmount/StakingAmountProtocols.swift
stas-litvinov: I meant checking amount on the next screen after main. If Amount + Fee is greater than available amount, we could just decrease it to be equal to max available amount or notify user with an alert that might look like "Amount you entered before is insufficient because of fee. Please, recalculate". Let's leave it as it is again and discuss later

stas-litvinov (approved):

PR #141, opened 2021-03-16
FLW-636 Network info view

stas-litvinov:
Add network view widget designs


fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -24,6 +25,16 @@ final class StakingMainPresenter {
         self.logger = logger
     }
 
+    private func provideChain() {
+        guard let viewModelFactory = networkInfoViewModelFactory else {
+            return
+        }
+
+        let chainModel = viewModelFactory.createChainViewModel()
```
ERussel, line 33: Let's just pass chain to view as it is and allow view to decide how to display. Because creating view model from a single title in a separate factory looks like overengineering.

fearless/Modules/Staking/StakingMain/StakingMainViewController.swift
```diff
@@ -55,7 +74,31 @@ final class StakingMainViewController: UIViewController, AdaptiveDesignable {
...
+        amountInputView.textField.resignFirstResponder()
+        let animationOptions: UIView.AnimationOptions = sender.isActivated ? .curveEaseIn : .curveEaseOut
+
+        self.storiesViewZeroHeightConstraint?.isActive = sender.isActivated
+        UIView.animate(
+            withDuration: 0.35,
+            delay: 0.0,
+            options: [animationOptions],
+            animations: {
+                self.view.layoutIfNeeded()
```
ERussel, line 87: Why do we need to call this twice?

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
stas-litvinov: Well, actually I was going to use the factory to provide network info data anyway. Wouldn't it be nice to keep all the methods inside?

fearless/Modules/Staking/StakingMain/StakingMainViewController.swift
stas-litvinov: You are right, the second one is not necessary

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
ERussel: Got it. Thanks

ERussel (approved):

PR #157, opened 2021-03-29
FLW-677 Implement "Payout confirmation" layout

ef1rspb:


fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationViewController.swift
```diff
@@ -0,0 +1,146 @@
+import UIKit
+import SoraFoundation
+
+final class StakingPayoutConfirmationViewController: UIViewController, ViewHolder {
+
+    typealias RootViewType = StakingPayoutConfirmationViewLayout
+
+    let presenter: StakingPayoutConfirmationPresenterProtocol
+    let localizationManager: LocalizationManagerProtocol?
```
ERussel, line 9: There is no need in the declaration because Localizable handles storing of this value. It also automatically adds observer when the value is set.

ERussel (approved):

PR #170, opened 2021-04-03
Integrate Codecov

ef1rspb:


pgolovkin:
Like it! Can we have the same for SORA?

ef1rspb:
> Like it! Can we have the same for SORA?

😏 Sure!

ERussel (approved):

PR #178, opened 2021-04-09
FLW-717 Add "Calculate reward payouts" test

ef1rspb:


fearlessIntegrationTests/RewardPayoutsServiceTests.swift
```diff
@@ -213,6 +213,236 @@ class RewardPayoutsServiceTests: XCTestCase {
...
+                            controller: controller,
+                            chain: chain,
+                            engine: engine,
+                            codingFactory: factory,
+                            queue: queue
+                        )
+                        .first!.value!
+                    }
+
+                let eras = Set<UInt32>(activeEra-84..<activeEra) // TODO use currentEra
```
ERussel, line 304: Let's make an exercise with current era too, so we don't miss this in the service

ERussel (approved):

PR #185, opened 2021-04-13
Fix election status

ERussel:
- subscribe to Staking.ElectionStatus if available otherwise try to subscribe to ElectionProviderMultiPhase.CurrentPhase

### DIFF
--- a/fearless/Common/Services/WebSocketService/StorageSubscription/ElectionStatusSubscription.swift
+++ b/fearless/Common/Services/WebSocketService/StorageSubscription/ElectionStatusSubscription.swift
@@ -14,6 +14,10 @@ final class ElectionStatusSubscription: WebSocketSubscribing {
     private var subscription: StorageChildSubscribing?
     private var subscriptionId: UInt16?
 
+    deinit {
+        unsubscribeRemote()
+    }
+
     init(
         engine: JSONRPCEngine,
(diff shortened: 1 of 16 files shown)


fearless/Common/Services/WebSocketService/StorageSubscription/ElectionStatusSubscription.swift
```diff
@@ -0,0 +1,140 @@
...
+import FearlessUtils
+
+final class ElectionStatusSubscription: WebSocketSubscribing {
+    let childSubscriptionFactory: ChildSubscriptionFactoryProtocol
+    let runtimeService: RuntimeCodingServiceProtocol
+    let operationManager: OperationManagerProtocol
+    let engine: JSONRPCEngine
+    let logger: LoggerProtocol?
+
+    private let mutex = NSLock()
```
stas-litvinov, line 12: To be more consistent I would suggest to call all locks the same way, either mutex or lock

stas-litvinov (approved):

fearless/Common/Services/WebSocketService/StorageSubscription/ElectionStatusSubscription.swift
ERussel: Thanks! Will fix in the separate PR

PR #186, opened 2021-04-14
FLW-656 Stake screen

stas-litvinov:
- Change screen title (Setup staking -> Stake)
- Remove "Amount to stake" caption
- Change Amount element (Balance -> Available)
- Change Reward destination caption (Choose reward destination -> Reward destination)
- Change selection indicators
- Change reward display (remove texts, add amount, add APY)
- Change Payout account element (Payout account -> Reward destination, Right arrow -> Down arrow)
- Change bottom sheet title (Payout account -> Reward destination)
- Change "learn more" text (payouts -> rewards)
- Change Next button (Next -> Continue)


fearless/Common/Extension/Foundation/NumberFormatter.swift
```diff
@@ -24,11 +24,17 @@ extension NumberFormatter {
     static var percentAPY: NumberFormatter {
         let numberFormatter = NumberFormatter.percent
         numberFormatter.minimumFractionDigits = 2
-        numberFormatter.positivePrefix = ""
+        numberFormatter.positivePrefix = "+"
         numberFormatter.percentSymbol = "% APY"
         return numberFormatter
     }
 
+    static var percentAPR: NumberFormatter {
```
ERussel, line 32: I suggest not to reuse APY implementation and just setup formatter from scratch

fearless/Common/View/RewardSelectionView.swift
```diff
@@ -3,10 +3,9 @@ import SoraUI
 
 class RewardSelectionView: BackgroundedContentControl {
     private(set) var titleLabel: UILabel!
-    private(set) var subtitleLabel: UILabel!
     private(set) var earningsTitleLabel: UILabel!
     private(set) var earningsSubtitleLabel: UILabel!
-    private(set) var iconView: UIImageView!
+    private(set) var iconView: UIButton!
```
ERussel, line 8: Why do we need button? Touch event is working on the whole view

fearless/Modules/Staking/StakingAmount/ViewModel/RewardDestinationViewModelFactory.swift
```diff
@@ -72,21 +72,24 @@ final class RewardDestinationViewModelFactory: RewardDestinationViewModelFactory
     ) -> LocalizableResource<RewardDestinationViewModelProtocol> {
         LocalizableResource { locale in
             let amountFormatter = amountFormatter.value(for: locale)
-            let percentageFormatter = NumberFormatter.percent.localizableResource().value(for: locale)
+            let percentageAPYFormatter = NumberFormatter.percentAPY.localizableResource().value(for: locale)
```
ERussel, line 75: I suggest to move NumberFormatter.percentAPY outside of the closure, so we just call value(for:) inside

fearless/Modules/Staking/StakingAmount/ViewModel/RewardDestinationViewModelFactory.swift
```diff
@@ -72,21 +72,24 @@ final class RewardDestinationViewModelFactory: RewardDestinationViewModelFactory
...
+            let percentageAPRFormatter = NumberFormatter.percentAPR.localizableResource().value(for: locale)
+
+            let zeroAmount = amountFormatter.string(from: 0.0) ?? "0"
```
ERussel, line 78: In the view we already handle empty state by assigning zeros. Why do we need it also here?

fearless/Common/View/RewardSelectionView.swift
stas-litvinov: Button provides states out of the box, so we don't need to change images manually and store state

fearless/Common/Extension/Foundation/NumberFormatter.swift
stas-litvinov: Okay, what about reusing percent implementation?

fearless/Modules/Staking/StakingAmount/ViewModel/RewardDestinationViewModelFactory.swift
stas-litvinov: Yes, we handle empty state, but here we try to handle case when `string(from:)` returns nil. Moreover, `DestinationReturnViewModel` does not accept Optionals and if we make zeroAmount Optional, the next code would look absolutely terrible:
`let rewardViewModel = DestinationReturnViewModel(
    restakeAmount: restakeAmount ?? zeroAmount,
    restakePercentage: restakePercentage ?? "",
    payoutAmount: payoutAmount ?? zeroAmount,
    payoutPercentage: payoutPercentage ?? ""
)`

fearless/Common/Extension/Foundation/NumberFormatter.swift
ERussel: Not great fan of it but at least it will be consistent with APY

fearless/Modules/Staking/StakingAmount/ViewModel/RewardDestinationViewModelFactory.swift
ERussel: Got it. Looks like we need an empty state for it like shimmering. Will discuss it later

ERussel (approved):

PR #190, opened 2021-04-16
FLW-733 Payouts list: Implement ErrorStateView

ef1rspb:
https://user-images.githubusercontent.com/17689921/115016066-421aeb00-9ebd-11eb-9064-9785b413e63c.mov


fearless/Common/Services/PayoutRewardsService/PayoutRewardsService.swift
```diff
@@ -166,12 +166,13 @@ final class PayoutRewardsService: PayoutRewardsServiceProtocol {
                 + prefsByEraWrapper.allOperations + [eraInfoOperation, payoutOperation]
 
             payoutOperation.completionBlock = {
-                do {
-                    let payouts = try payoutOperation.extractNoCancellableResultData()
-                    completion(.success(payouts))
-                } catch {
-                    completion(.failure(error))
-                }
+                completion(.failure(PayoutError.unknown))
```
ef1rspb, line 169: delete

fearless/Common/Protocols/ErrorStateViewProtocol.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
+
+protocol ErrorStateViewProtocol: AnyObject {
```
ERussel, line 3: I think we can just return ErrorView from ```viewForEmptyState``` delegate method of EmptyStateView. And also we can set ViewController as delegate for ErrorView and call presenter ```reload``` function in delegate method. I think It will be more clear.

Currently, imagine a situation when someone comes to fix something in view. And it will be hard to understand that this tap on error view is actually handled not by the controller but by closure in the bottom of the presenter :)

fearless/Common/Services/PayoutRewardsService/PayoutRewardsService.swift
```diff
@@ -178,11 +178,12 @@ final class PayoutRewardsService: PayoutRewardsServiceProtocol {
                 + [eraInfoOperation, payoutOperation]
 
             payoutOperation.completionBlock = {
+                // completion(.failure(PayoutError.unknown))
```
ef1rspb, line 181: delete

fearless/Common/Services/PayoutRewardsService/PayoutRewardsServiceProtocol.swift
```diff
@@ -7,6 +7,10 @@ protocol PayoutRewardsServiceProtocol {
     func fetchPayoutRewards(completion: @escaping PayoutRewardsClosure)
 }
 
-enum PayoutError: Error {
+enum PayoutError: LocalizedError {
```
ERussel, line 10: We use ErrorContentConvertible protocol for this. Once error adopted it can be displayed automatically through ErrorPresentable protocol

ERussel (approved):

ERussel (approved):

PR #200, opened 2021-04-21
Validator payout service

ERussel:
- add logic to create payout validator service when validator stash is passed to module
- add logic to automatically break list of keys into pages too query storage


fearless/Common/Operation/StorageRequestFactory.swift
```diff
@@ -84,29 +86,57 @@ final class StorageRequestFactory: StorageRequestFactoryProtocol {
         }
     }
 
+    private func createQueryOperation(
+        for keys: @escaping () throws -> [Data],
+        engine: JSONRPCEngine
+    ) -> BaseOperation<[[StorageUpdate]]> {
+        OperationCombiningService<[StorageUpdate]>(
```
ef1rspb, line 93:
```swift
OperationCombiningService<[StorageUpdate]>(operationManager: operationManager) {
```
or 
```swift
OperationCombiningService<[StorageUpdate]>(
    operationManager: operationManager
) {
```

ef1rspb (approved):

fearless/Common/Operation/StorageRequestFactory.swift
ERussel: will fix in the next PR

PR #205, opened 2021-04-21
FLW-768 Add "Staking balance" to Manage bottom sheet from Staking screen

ef1rspb:
Screenshot 2021-04-15 at 18 48 37


fearless/Modules/Staking/StakingManage/View/StakingManageCell.swift
```diff
@@ -0,0 +1,61 @@
...
+        contentView.addSubview(iconImageView)
+        iconImageView.snp.makeConstraints { make in
+            make.leading.equalToSuperview().inset(UIConstants.horizontalInset)
+            make.centerY.equalToSuperview()
+            make.size.equalTo(24)
+        }
+
+        contentView.addSubview(titleLabel)
+        titleLabel.snp.makeConstraints { make in
+            make.leading.equalTo(iconImageView.snp.trailing).offset(15)
```
ERussel, line 45: we usually use 16. Is this from mockups?

fearless/Modules/Staking/StakingManage/View/StakingManageCell.swift
```diff
@@ -0,0 +1,61 @@
...
+            make.bottom.top.equalToSuperview().inset(15)
```
ERussel, line 46: Why not to put centerY? Then you don't need to hardcode insets

ERussel (approved):

fearless/Modules/Staking/StakingManage/View/StakingManageCell.swift
ef1rspb: mockups are currently in progress, so spacing is between 13-15 :(

fearless/Modules/Staking/StakingManage/View/StakingManageCell.swift
ef1rspb: because the layout will shrink to either iconImageView height or titleLabel intrinsic height

PR #206, opened 2021-04-22
Payouts refactoring

ERussel:
- add fee formatting on the confirm screen
- fix price offsets
- fix account view highlighted state
- fix localizatioin


stas-litvinov:
I noticed one more thing: on both payout details and payout confirmation screens spacing between account element and the next string is less than designed
image

ERussel:
> I noticed one more thing: on both payout details and payout confirmation screens spacing between account element and the next string is less than designed
> image

fixed

stas-litvinov (approved):

PR #217, opened 2021-04-30
FLW-770 "Bond more" module

ef1rspb:
Merged branch with Unbond flow, so don't review this stuff. 
Tests are coming..


fearlessTests/Mocks/ViewModel/StubBalanceViewModelFactory.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+import SoraFoundation
+import CommonWallet
+@testable import fearless
+
+struct StubBalanceViewModelFactory: BalanceViewModelFactoryProtocol {
```
ERussel, line 6: Why do we need stub the factory? We can run to the situation when we receive some broken data from interactor but doesn't detect this because the factory is stubbed.

ef1rspb: In this test case I just want to test "continue" action, so balance factory is unnecessary and I used stub. In the next tests I'll use real implementation.

ERussel (approved):

PR #219, opened 2021-05-04
FLW-782 Unbond flow: Unbond screen

ERussel:
- reusable code for subscriptions moved to separate extension
- add validation runner to simplify data validation
- add extrinsic proxy to simplify fee retrieval with cache support
- add unbond setup screen


fearless/Modules/Staking/StakingUnbondSetup/StakingUnbondSetupViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        localizationManager: LocalizationManagerProtocol
+    ) {
+        self.presenter = presenter
+
+        super.init(nibName: nil, bundle: nil)
+
+        self.localizationManager = localizationManager
+    }
+
+    var uiFactory: UIFactoryProtocol = UIFactory()
```
stas-litvinov, line 25: private let?

fearless/Modules/Staking/StakingUnbondSetup/StakingUnbondSetupViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    init(
+        presenter: StakingUnbondSetupPresenterProtocol,
+        localizationManager: LocalizationManagerProtocol
+    ) {
+        self.presenter = presenter
+
+        super.init(nibName: nil, bundle: nil)
+
+        self.localizationManager = localizationManager
```
stas-litvinov, line 22: Why does this particular order matter?

fearless/Common/View/TitleValueView.swift
```diff
@@ -0,0 +1,56 @@
...
+
+        addSubview(titleLabel)
+        titleLabel.snp.makeConstraints { make in
+            make.leading.centerY.equalToSuperview()
+        }
+
+        addSubview(valueLabel)
+        valueLabel.snp.makeConstraints { make in
+            make.trailing.centerY.equalToSuperview()
+            make.leading.greaterThanOrEqualTo(titleLabel.snp.trailing).offset(8.0)
```
stas-litvinov, line 53: I suggest to move all insets to local Constants as we do in other views

fearless/Common/View/UIFactory.swift
```diff
@@ -210,6 +212,53 @@ final class UIFactory: UIFactoryProtocol {
...
+        amountInputView.balanceLabel.textColor = R.color.colorLightGray()
+        amountInputView.balanceLabel.font = .p2Paragraph
+        amountInputView.textField.font = .h4Title
+        amountInputView.textField.tintColor = R.color.colorWhite()
+        amountInputView.verticalSpacing = 2.0
+        amountInputView.iconRadius = 12.0
+        amountInputView.contentInsets = UIEdgeInsets(
+            top: 8.0,
+            left: UIConstants.horizontalInset,
+            bottom: 8.0,
```
stas-litvinov, line 245: I suggest to move all insets to local Constants as we do in other views

fearless/Common/Validation/Validators/WarningConditionViolation.swift
```diff
@@ -0,0 +1,24 @@
...
+        self.checkCondition = checkCondition
+        self.onWarning = onWarning
+    }
+
+    func validate(notifying delegate: DataValidatingDelegate) -> DataValidationProblem? {
+        if checkCondition() {
+            return nil
+        }
+
+        onWarning(delegate)
```
stas-litvinov, line 20: I would rewrite it to make look more natural:
```
if violatesCondition() {
    notify(delegate)
    return .warning
}
```
or maybe  
`if satisfiesCondition() {`

stas-litvinov (approved):

fearless/Modules/Staking/StakingUnbondSetup/StakingUnbondSetupViewController.swift
ERussel: The purpose here is to override the property if we will need in future

fearless/Modules/Staking/StakingUnbondSetup/StakingUnbondSetupViewController.swift
ERussel: The problem here is that ```presenter``` is local variable so it must be initialised before calling super. ```LocalizationManager``` is requirement of the protocol with default value and it must be initialised after super

fearless/Common/View/TitleValueView.swift
ERussel: So we are moving to code based layout and suggestion here is that we don't move constants to the header because it is used in one place with clear naming

fearless/Common/View/UIFactory.swift
ERussel: So we are moving to code based layout and suggestion here is that we don't move constants to the header because it is used in one place with clear naming

fearless/Common/Validation/Validators/WarningConditionViolation.swift
ERussel: Will fix in next PR

PR #220, opened 2021-05-04
FLW-786 Unbond flow: Confirmation screen

ERussel:
- add unbond confirmation screen


fearless/Modules/Staking/StakingUnbondConfirm/StakingUnbondConfirmPresenter.swift
```diff
@@ -0,0 +1,289 @@
...
+
+    func refreshFeeIfNeeded() {
+        guard fee == nil, controller != nil, payee != nil, bonded != nil, minimalBalance != nil else {
+            return
+        }
+
+        interactor.estimateFee(for: inputAmount, resettingRewardDestination: shouldResetRewardDestination)
+    }
+
+    init(
```
stas-litvinov, line 84: It's rather unusual to see an initialiser after public func

fearless/Modules/Staking/StakingUnbondConfirm/StakingUnbondConfirmInteractor.swift
```diff
@@ -0,0 +1,241 @@
...
+    let singleValueProviderFactory: SingleValueProviderFactoryProtocol
+    let substrateProviderFactory: SubstrateDataProviderFactoryProtocol
+    let settings: SettingsManagerProtocol
+    let runtimeService: RuntimeCodingServiceProtocol
+    let operationManager: OperationManagerProtocol
+    let extrinsicServiceFactory: ExtrinsicServiceFactoryProtocol
+    let feeProxy: ExtrinsicFeeProxyProtocol
+    let accountRepository: AnyDataProviderRepository<AccountItem>
+    let chain: Chain
+    let assetId: WalletAssetId
```
stas-litvinov, line 19: Looks like everything here should be private

fearless/Common/View/UIFactory.swift
```diff
@@ -43,9 +43,21 @@ protocol UIFactoryProtocol {
...
+
+    func createNetworkFeeConfirmView() -> NetworkFeeConfirmView
+
+    func createTitleValueView() -> TitleValueView
+
+    func createHintView() -> HintView
 }
 
 final class UIFactory: UIFactoryProtocol {
+    static let `default` = UIFactory()
```
stas-litvinov, line 59: Instead of using reserved word I would suggest to rename it to defaultFactory

stas-litvinov (approved):

fearless/Modules/Staking/StakingUnbondConfirm/StakingUnbondConfirmInteractor.swift
ERussel: We just need a rule for this. In my opinion it is convenient to have public properties which just were passed as parameters in init.

fearless/Common/View/UIFactory.swift
ERussel: But think about usage: ```UIFactory.default``` vs ```UIFactory.defaultFactory```

PR #230, opened 2021-05-12
FLW-813 Update available connections on Networks screen

ef1rspb:


fearless/Common/Storage/Migration/NetworkConnectionsMigrator.swift
```diff
@@ -0,0 +1,16 @@
...
+final class NetworkConnectionsMigrator: Migrating {
+    private(set) var settings: SettingsManagerProtocol
+
+    init(settings: SettingsManagerProtocol) {
+        self.settings = settings
+    }
+
+    func migrate() throws {
+        let selectedConnection = settings.selectedConnection
+        if ConnectionItem.deprecatedConnections.contains(selectedConnection) {
```
ERussel, line 12: This doesn't take into account connection type. We should get type from selected connection and find first connection with the same type which is available. Let's also add test for this case ;)

fearless/Common/Storage/Migration/NetworkConnectionsMigrator.swift
```diff
@@ -0,0 +1,20 @@
...
+        self.settings = settings
+    }
+
+    func migrate() throws {
+        let selectedConnection = settings.selectedConnection
+        let deprecatedConnections = ConnectionItem.deprecatedConnections
+        let supportedConnections = ConnectionItem.supportedConnections
+
+        if deprecatedConnections.contains(selectedConnection) {
+            let suitableConnection = supportedConnections.first(where: { $0.type == selectedConnection.type })
```
ERussel, line 16: I suggest do nothing if we didn't find suitable connection. We can't use default one because it can be from different network.

ERussel (approved):

PR #233, opened 2021-05-13
FLW-702 Alerts widget UI

ef1rspb:
Show Alerts widget with stub data

### DIFF
--- a/fearless/Common/Extension/UIKit/UIView+Separator.swift
+++ b/fearless/Common/Extension/UIKit/UIView+Separator.swift
@@ -6,4 +6,29 @@ extension UIView {
         view.backgroundColor = color
         return view
     }
+
+    static func createSeparator(
+        color: UIColor? = R.color.colorLightGray(),
+        horizontalInset: CGFloat = 0
+    ) -> UIView {
+        SeparatorWithHorizontalInset(color: color, horizontalInset: horizontalInset)
+    }
+}
+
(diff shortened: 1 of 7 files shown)


fearless/Common/Extension/UIKit/UIView+Separator.swift
```diff
@@ -6,4 +6,29 @@ extension UIView {
         view.backgroundColor = color
         return view
     }
+
+    static func createSeparator(
```
ERussel, line 10: I think we already have a method in UIFactory

fearless/Modules/Staking/StakingMain/View/AlertsView.swift
```diff
@@ -0,0 +1,228 @@
...
+            noAlertsLabel.isHidden = true
+            alertsStackView.isHidden = false
+
+            let itemViews = alerts.map { alert -> UIView in
+                let itemView = AlertItemView(stakingAlert: alert, locale: locale)
+                itemView.addTarget(self, action: #selector(handleSelectItem), for: .touchUpInside)
+                return itemView
+            }
+
+            let separators = (0 ..< itemViews.count).map { _ -> UIView in
```
ERussel, line 103: I think we could have BorderedContainerView (appearance of the separator is controlled via borderType) inside AlertItemView

ERussel (approved):

PR #235, opened 2021-05-14
FLW-818 iOS: Recommended Validators: Add only up to 2 nodes from 1 cluster

ERussel:
- limit up to 2 validators from the single cluster in recommendations


antonkhvorov:
We should also represent that in the UI as another criteria which we are applying for creating Recommended list:
...
✅ Limit of 2 validators per identity

ef1rspb (approved):

PR #238, opened 2021-05-14
Fix migration issue

ERussel:
- clear selected account when there is no corresponding keypair locally


fearlessTests/Common/Migration/InconsistentStateMigrationTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        try AccountCreationHelper.createAccountFromMnemonic(cryptoType: .sr25519,
+                                                            networkType: chain,
+                                                            keychain: keychain,
+                                                            settings: settings)
+
+        XCTAssertNotNil(settings.selectedAccount)
+
+        let migrator = InconsistentStateMigrator(
+            settings: settings,
+            keychain: InMemoryKeychain()
```
ef1rspb, line 24: keychain is defined at line 11

ef1rspb (approved):

fearlessTests/Common/Migration/InconsistentStateMigrationTests.swift
ERussel: Yes, this is done in purpose to immitate inconsistent state

PR #245, opened 2021-05-17
FLW-721 Alert: Tokens are available for Withdraw

ef1rspb:


fearless/Modules/Staking/StakingMain/ViewModel/StakingStateViewModelFactory+Alerts.swift
```diff
@@ -44,7 +58,21 @@ extension StakingStateViewModelFactory {
     func stakingAlertsForValidatorState(_ state: ValidatorState) -> [StakingAlert] {
         switch state.status {
         case .active:
-            return []
+            guard
```
ef1rspb, line 61: copypaste(( maybe define ledgerInfo in commonData?

fearless/Modules/Staking/StakingMain/ViewModel/StakingStateViewModelFactory+Alerts.swift
```diff
@@ -5,7 +5,21 @@ extension StakingStateViewModelFactory {
     func stakingAlertsForNominatorState(_ state: NominatorState) -> [StakingAlert] {
         switch state.status {
         case .active:
-            return []
+            guard
```
ERussel, line 8: Redeemable alert should not depend on staking status. Because of the fact that this information is stored in StakingLedger and not in Nomination then tokens can be redeemed anytime unless it is election period.

fearless/Modules/Staking/StakingMain/ViewModel/StakingStateViewModelFactory+Alerts.swift
ERussel: We can move this into separate function because it depends on era and ledger. This info available in nominator, validator and bonded states.

fearless/Modules/Staking/StakingMain/ViewModel/StakingStateViewModelFactory+Alerts.swift
```diff
@@ -3,71 +3,59 @@ import SoraFoundation
...
-            } else {
-                return [.nominatorNoValidators]
-            }
-        case .waiting:
-            return []
-        case .undefined:
-            return []
-        }
+        [
+            findElectionAlert(commonData: state.commonData),
```
ERussel, line 7: Why other alerts gone?

ERussel (approved):

PR #246, opened 2021-05-17
FLW-807 Fix Incoming transaction detail: total amount

ef1rspb:


ERussel (changes requested):
Let's just don't include fee view model for incoming transfer

ERussel (approved):

PR #250, opened 2021-05-18
FLW-832 iOS. Wrong existential deposit in send flow for Kusama network.

stas-litvinov:
change existential deposit for Kusama network


antonkhvorov (changes requested):
You need to obtain this from Runtime, not to fix the hard coded value. You can check how this is obtained on Staking screen, for minimum value to stake.

PR #252, opened 2021-05-19
FLW-678 iOS. Staking. The back button hasn't offset.

stas-litvinov:
change back button appearance in Staking modules
(back button appearance in Wallet modules should be implemented as separate issue)


ERussel (changes requested):
I think we need to update the Capital at the same time. I think we can add optional back button icon to the WalletNavigationBarStyleProtocol.

ef1rspb (approved):

fearless/Common/ViewController/NavigationController/FearlessNavigationController.swift
```diff
@@ -32,6 +32,10 @@ final class FearlessNavigationController: UINavigationController, UINavigationCo
 
         navigationBar.setBackgroundImage(UIImage(), for: UIBarMetrics.default)
         navigationBar.shadowImage = UIImage()
+        navigationBar.backIndicatorImage = R.image.iconBack()
+        navigationBar.backIndicatorTransitionMaskImage = R.image.iconBack()
+
+        navigationItem.leftItemsSupplementBackButton = true
```
ERussel, line 38: Navigation item is bonded to concrete controller. Does this work when we push new controller to the stack?

stas-litvinov: Yes, it works like a charm
image
image

ERussel: I tried to run and don't see any difference with this line of code ```navigationItem.leftItemsSupplementBackButton = true``` and without it. As expected navigationItem of the UINavigationController shouldn't make any effect to the navigation's bar items. Can you any difference? May be I am missing something

ERussel (approved):

PR #254, opened 2021-05-19
FLW-837 Controller account: check balance is not zero

ef1rspb:

### DIFF
--- a/fearless/Modules/Staking/ControllerAccount/ControllerAccountInteractor.swift
+++ b/fearless/Modules/Staking/ControllerAccount/ControllerAccountInteractor.swift
@@ -23,6 +23,7 @@ final class ControllerAccountInteractor {
 
     private var stashItemProvider: StreamableProvider<StashItem>?
     private var accountInfoProvider: AnyDataProvider<DecodedAccountInfo>?
+    private var controllerAccountInfoProvider: AnyDataProvider<DecodedAccountInfo>?
     private var ledgerProvider: AnyDataProvider<DecodedLedgerInfo>?
     private var extrinsicService: ExtrinsicServiceProtocol?
 
@@ -78,6 +79,14 @@ extension ControllerAccountInteractor: ControllerAccountInteractorInputProtocol 
         } catch {
             presenter.didReceiveFee(result: .failure(error))
         }
+    }
+
+    func fetchControllerAccountInfo(controllerAddress: AccountAddress) {
+        clear(dataProvider: &controllerAccountInfoProvider)
+        controllerAccountInfoProvider = subscribeToAccountInfoProvider(
+            for: controllerAddress,
+            runtimeService: runtimeService
+        )
     }
 
     func fetchLedger(controllerAddress: AccountAddress) {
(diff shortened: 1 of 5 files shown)


fearless/Modules/Staking/ControllerAccount/ControllerAccountInteractor.swift
```diff
@@ -80,6 +81,14 @@ extension ControllerAccountInteractor: ControllerAccountInteractorInputProtocol
         }
     }
 
+    func fetchControllerAccountInfo(controllerAddress: AccountAddress) {
+        clear(dataProvider: &controllerAccountInfoProvider)
+        controllerAccountInfoProvider = subscribeToAccountInfoProvider(
```
ERussel, line 86: One this which is I just noticed is that subscription just retrieves data from database. We need here the same approach as for ledger - just fetch from blockchain

ERussel (changes requested):
Replace account info subscription for controller with manual fetching from blockchain

fearless/Modules/Staking/ControllerAccount/ControllerAccountInteractor.swift
```diff
@@ -80,6 +80,24 @@ extension ControllerAccountInteractor: ControllerAccountInteractorInputProtocol
...
     }
 
+    func fetchControllerAccountInfo(controllerAddress: AccountAddress) {
+        do {
+            let accountId = try addressFactory.accountId(fromAddress: controllerAddress, type: chain.addressType)
+
+            let accountInfoOperation = createAccountInfoFetchOperation(accountId)
+            accountInfoOperation.targetOperation.completionBlock = { [weak presenter] in
+                DispatchQueue.main.async {
+                    if let result = accountInfoOperation.targetOperation.result {
```
ERussel, line 90: I would suggest to use ```extractNoCancellableResultData``` or please handle ```else```

fearless/Modules/Staking/ControllerAccount/ControllerAccountPresenter.swift
```diff
@@ -62,13 +63,14 @@ final class ControllerAccountPresenter {
         }
     }
 
-    func refreshLedgerIfNeeded() {
+    private func refreshControllerInfoIfNeeded() {
         guard let chosenControllerAddress = chosenAccountItem?.address else {
             return
         }
         if chosenControllerAddress != stashItem?.controller {
             stakingLedger = nil
```
ERussel, line 71: Should we assign ```nil``` also to ```controllerBalance``` ?

fearless/Modules/Staking/ControllerAccount/ControllerAccountPresenter.swift
```diff
@@ -139,6 +141,7 @@ extension ControllerAccountPresenter: ControllerAccountPresenterProtocol {
                 fee: fee,
                 locale: locale
             ),
+            dataValidatingFactory.controllerBalanceIsNotZero(controllerBalance, locale: locale),
```
ERussel, line 144: Le't also refresh in case of error like we do with fee, so we could handle cases when user adds tokens to the controller and then tries to proceed

ERussel (approved):

PR #258, opened 2021-05-20
FLW-870 Show "missing controller" error on Redeem alert action

ef1rspb:

### DIFF
--- a/fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
+++ b/fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
@@ -261,11 +261,8 @@ extension StakingMainPresenter: StakingMainPresenterProtocol {
 }
 
 extension StakingMainPresenter: StakingStateMachineDelegate {
-    func stateMachineDidChangeState(_ stateMachine: StakingStateMachineProtocol) {
+    func stateMachineDidChangeState(_: StakingStateMachineProtocol) {
         provideState()
-        if let stashItem = stateMachine.viewState(using: { (state: BondedState) in state.stashItem }) {
-            interactor.fetchController(for: stashItem.controller)
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -211,13 +245,27 @@ extension StakingMainPresenter: StakingMainPresenterProtocol {
...
+        }
         wireframe.showRedeem(from: view)
     }
 }
 
 extension StakingMainPresenter: StakingStateMachineDelegate {
-    func stateMachineDidChangeState(_: StakingStateMachineProtocol) {
+    func stateMachineDidChangeState(_ stateMachine: StakingStateMachineProtocol) {
         provideState()
+        if let stashItem = stateMachine.viewState(using: { (state: BondedState) in state.stashItem }) {
```
ERussel, line 266: This won't work for all cases, for example when we are nominator. I would rather bind the logic when we receive stash item. Ideally, I would store stashAccount and controllerAccount in state.

ef1rspb: oops, this is redundant code, i've removed those lines. Currently fetchController calls on stashItem
Screenshot 2021-05-20 at 14 03 08

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -95,16 +96,49 @@ extension StakingMainPresenter: StakingMainPresenterProtocol {
...
+                let controllerAddress = bonded.stashItem.controller
+                wireframe.presentMissingController(from: view, address: controllerAddress, locale: locale)
             }
-        } else {
-            wireframe.showSetupAmount(from: view, amount: amount)
+
+            return
+        }
+
+        let optExistingBonding: ExistingBonding? = stateMachine.viewState { (state: BondedState) in
```
ERussel, line 114: Looks like we already have state in line 99 so we don't need to query state machine one more time

ef1rspb: 🤦‍♂️ sure!

ERussel (approved):

PR #259, opened 2021-05-20
FLW-868 Make OnFinality nodes default one

ef1rspb:


fearless/Common/Extension/Storage/Storage+Identifiable.swift
```diff
@@ -6,7 +6,7 @@ extension AccountItem: Identifiable {
 }
 
 extension ConnectionItem: Identifiable {
-    var identifier: String { url.absoluteString }
+    var identifier: String { (url.scheme ?? "") + "://" + (url.host ?? "") }
```
antonkhvorov, line 9: what about IPs? I think that's the reason why before it was an absoluteString

fearless/Common/Extension/Model/ConnectionItem+Default.swift
```diff
@@ -4,22 +4,22 @@ import IrohaCrypto
 extension ConnectionItem {
     static var defaultConnection: ConnectionItem {
```
antonkhvorov, line 5: Do we still need that? @ERussel 
Before we had a Kusama account as a default, thus this was probably used, but now users can create an account in any network as their first account, so maybe we can just use supportedConnections?

fearless/Common/Extension/Storage/Storage+Identifiable.swift
ERussel: Didn't get why we need to change that?

fearless/Common/Extension/Model/ConnectionItem+Default.swift
ERussel: Yep, let's leave as it is for now

ERussel (approved):

PR #261, opened 2021-05-20
FLW-697 Remove "active era" label from "Your stake" widget

ef1rspb:


stas-litvinov (approved):

antonkhvorov (changes requested):
no, we need this info, blame designers for removing it from the mockups 😏

PR #262, opened 2021-05-21
FLW-484 Send from account with only Existential deposit

ef1rspb:


fearless/Modules/Wallet/Transfer/TransferValidatingError.swift
```diff
@@ -3,6 +3,7 @@ import CommonWallet
 
 enum FearlessTransferValidatingError: Error {
     case receiverBalanceTooLow
+    case senderBalanceTooLow
```
ERussel, line 6: I would prefer more specific: ```cantPayFee```

fearless/en.lproj/Localizable.strings
```diff
@@ -406,3 +406,4 @@ Derivation path: %@";
 "staking.earned" = "Earned";
 "staking.controller.account.zero.balance" = "We found that this account has no free tokens, are you sure that you want to change the controller?";
 "wallet.send.balance.minimal" = "Minimal balance";
+"wallet.fee.over.existential.deposit" = "You have to pay a fee over and above the existential deposit.";
```
ERussel, line 409: You should have funds to pay fee and remain above minimal balance

ERussel (approved):
Separate respect for adding tests! Thank you!

PR #294, opened 2021-06-22
FLW-892 Selected Validators screen

stas-litvinov:
Selected validators scene implementation: 
- List with all validators available, with their names/addresses, amount, rewards
- Switch to edit mode to remove any validator from the list
- Continue to Confirm screen on "Continue" tap
image


fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListPresenter.swift
```diff
@@ -0,0 +1,109 @@
...
+        )
+        wireframe.present(validatorInfo, from: view)
+    }
+
+    func removeItem(at index: Int) {
+        let validator = selectedValidatorList[index]
+        delegate?.didRemove(validator)
+
+        selectedValidatorList.remove(at: index)
+        let viewModel = createViewModel()
```
ERussel, line 77: Why do we need to store view model is we always recreate them?

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListPresenter.swift
```diff
@@ -0,0 +1,109 @@
...
+    func didSelectValidator(at index: Int) {
+        let validatorInfo = selectedValidatorInfo(
+            from: selectedValidatorList[index]
+        )
+        wireframe.present(validatorInfo, from: view)
+    }
+
+    func removeItem(at index: Int) {
+        let validator = selectedValidatorList[index]
+        delegate?.didRemove(validator)
```
ERussel, line 74: Actually we haven't removed it yet :) may be call it before or after updating view

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+    private func updateProceedButton() {
+        let cellViewModels = viewModel?.cellViewModels ?? []
+
+        let buttonTitle: String
+        let enabled: Bool
+        let fillColor: UIColor
+
+        if cellViewModels.count > selectedValidatorsLimit {
+            enabled = false
+            fillColor = R.color.colorDarkGray()!
```
ERussel, line 101: I think we should refactor our ```applyDisabledStyle``` function for the TriangularedButton

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+            fillColor = R.color.colorAccent()!
+            buttonTitle = R.string.localizable
+                .commonContinue(
+                    preferredLanguages: selectedLocale.rLanguages
+                )
+        }
+
+        rootView.proceedButton.triangularedView?.fillColor = fillColor
+        rootView.proceedButton.imageWithTitleView?.title = buttonTitle
+        rootView.proceedButton.isEnabled = enabled
```
ERussel, line 119: I think it is also changes alpha for disabled state and it not 1 by default. So it is still better to refactor ```applyDisabledStyle```

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+// MARK: - SelectedValidatorListViewProtocol
+
+extension SelectedValidatorListViewController: SelectedValidatorListViewProtocol {
+    func updateViewModel(_ viewModel: SelectedValidatorListViewModel) {
+        self.viewModel = viewModel
+        updateHeaderView()
+        updateProceedButton()
+    }
+
+    func reload(_ viewModel: SelectedValidatorListViewModel) {
```
ERussel, line 174: I would suggest to rename to ```didReload```

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+            updateProceedButton()
+            updateHeaderView()
+        }
+    }
+}
+
+// MARK: - SelectedValidatorListViewProtocol
+
+extension SelectedValidatorListViewController: SelectedValidatorListViewProtocol {
+    func updateViewModel(_ viewModel: SelectedValidatorListViewModel) {
```
ERussel, line 168: I didn't saw that it is called at some point when reload is not called. So, can we combine them?

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+        self.viewModel = viewModel
+        updateProceedButton()
+        rootView.tableView.reloadData()
+    }
+
+    func didRemoveItem(at index: Int) {
+        let indexPath = IndexPath(row: index, section: 0)
+        rootView.tableView.deleteRows(at: [indexPath], with: .left)
+
+        let cellViewModels = viewModel?.cellViewModels ?? []
```
ERussel, line 184: So we are making an assumption that updateViewModel is called before this method. I would recommend to rename the method to didChangeViewModel(..., byRemovingItemAt index:).

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
```diff
@@ -0,0 +1,240 @@
...
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        presenter.didSelectValidator(at: indexPath.row)
+    }
+
+    func tableView(_ tableView: UITableView, viewForHeaderInSection _: Int) -> UIView? {
+        guard let viewModel = viewModel else { return nil }
+
+        let headerView: SelectedValidatorListHeaderView = tableView.dequeueReusableHeaderFooterView()
+        self.headerView = headerView
```
ERussel, line 232: I think it its unexpected behaviour. If we want to reload only the header of the section then we could get it through ```headerViewForSection:```.

ERussel (approved):

fearless/Modules/Staking/SelectValidatorsFlow/SelectedValidatorList/SelectedValidatorListViewController.swift
stas-litvinov: It's called whenever we remove a row. In this case we don't want to reload tableView since it will cause data inconsistency

PR #300, opened 2021-06-25
remove election period from staking

ERussel:


stas-litvinov (changes requested):
I would suggest to look at ```createElectionStatus(...)``` function in NominatorState+Status.swift and ValidatorState+Status.swift files. Looks like we don't need it

stas-litvinov (approved):

PR #303, opened 2021-07-01
FLW-1040 Save state of Network information widget

ef1rspb:
https://user-images.githubusercontent.com/17689921/124130975-02a97700-da88-11eb-8c0f-91b58e32919a.mov

### DIFF
--- a/fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
+++ b/fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
@@ -15,6 +15,7 @@ final class StakingMainPresenter {
 
     private var stateViewModelFactory: StakingStateViewModelFactoryProtocol
     private var stateMachine: StakingStateMachineProtocol
+    private var networkInfoViewIsExpanded: Bool?
 
     var chain: Chain? {
         stateMachine.viewState { (state: BaseStakingState) in state.commonData.chain }
@@ -66,7 +67,8 @@ final class StakingMainPresenter {
                     with: networkStakingInfo,
                     chain: chain,
                     minNominatorBond: commonData?.minNominatorBond,
-                    priceData: commonData?.price
+                    priceData: commonData?.price,
+                    viewIsExpanded: networkInfoViewIsExpanded ?? true
                 )
             view?.didRecieveNetworkStakingInfo(viewModel: networkStakingInfoViewModel)
         } else {
@@ -298,7 +300,7 @@ extension StakingMainPresenter: StakingMainPresenterProtocol {
     }
 
     func networkInfoViewDidChangeExpansion(isExpanded: Bool) {
-        networkInfoViewModelFactory.viewDidChangeExpansion(isExpanded: isExpanded)
+        interactor.saveNetworkInfoViewExpansion(isExpanded: isExpanded)
     }
 }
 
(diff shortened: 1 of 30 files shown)


fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -296,6 +296,10 @@ extension StakingMainPresenter: StakingMainPresenterProtocol {
         }
         wireframe.showRedeem(from: view)
     }
+
+    func networkInfoViewDidCollapsed(_ value: Bool) {
+        networkInfoViewModelFactory.viewDidCollapsed(value)
```
stas-litvinov, line 301: Function name is a little bit confusing to me. I suggest to rename it to something like networkInfoViewDidChangeExpansion(isExpanded: Bool)

fearless/Modules/Staking/StakingMain/StakingMainViewController.swift
```diff
@@ -343,6 +343,10 @@ extension StakingMainViewController: NetworkInfoViewDelegate {
     func animateAlongsideWithInfo(view _: NetworkInfoView) {
         scrollView.layoutIfNeeded()
     }
+
+    func viewCollapsed(_ value: Bool, view _: NetworkInfoView) {
+        presenter.networkInfoViewDidCollapsed(value)
```
stas-litvinov, line 348: Why do we need view parameter here?

fearless/Modules/Staking/StakingMain/View/NetworkInfoView.swift
```diff
@@ -97,6 +98,7 @@ final class NetworkInfoView: UIView {
         minimumStakeFiatAmountLabel.text = localizedViewModel.minimalStake?.price
         activeNominatorsLabel.text = localizedViewModel.activeNominators
         lockUpPeriodLabel.text = localizedViewModel.lockUpPeriod
+        setExpanded(!localizedViewModel.viewIsCollapsed, animated: false)
```
stas-litvinov, line 101: Looks like it is better to rename viewIsCollapsed to viewIsExpanded to make code more consistent and readable

fearless/Modules/Staking/StakingMain/StakingMainViewController.swift
ef1rspb: jus for future needs, it's common for delegate to pass object. You know 'func tableView(_ tableView: UITableView, did...)'

fearless/Modules/Staking/StakingMain/ViewModel/NetworkInfoViewModelFactory.swift
```diff
@@ -116,6 +125,10 @@ extension NetworkInfoViewModelFactory: NetworkInfoViewModelFactoryProtocol {
         }
     }
 
+    func viewDidChangeExpansion(isExpanded: Bool) {
+        settings.set(value: isExpanded, for: Self.viewIsExpandedKey)
```
ERussel, line 129: Ideally this logic should be in interactor

fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -66,7 +67,8 @@ final class StakingMainPresenter {
                     with: networkStakingInfo,
                     chain: chain,
                     minNominatorBond: commonData?.minNominatorBond,
-                    priceData: commonData?.price
+                    priceData: commonData?.price,
+                    viewIsExpanded: networkInfoViewIsExpanded ?? true
```
ERussel, line 71: I would put this just as separate method in view protocol, because it doesn't depend on data. Imagine situation when one has slow connection and open's the screen. The settings will not be applied until network data received

ERussel (approved):

PR #308, opened 2021-07-02
FLW-893 Search/filter validator by name/address

stas-litvinov:
- add online wannabe validator search
- update tests


fearless/Modules/Staking/Operations/ValidatorOperationFactory.swift
```diff
@@ -172,6 +176,49 @@ final class ValidatorOperationFactory {
         )
     }
 
+    private func createValidatorPrefsWrapper(for accountIdList: [AccountId])
+        -> CompoundOperationWrapper<[AccountAddress: ValidatorPrefs]> {
+        let addressType = chain.addressType
+
+        let runtimeFetchOperation = runtimeService.fetchCoderFactoryOperation()
+
+        let fetchOperation: CompoundOperationWrapper<[StorageResponse<ValidatorPrefs>]> =
```
ERussel, line 185: Don't see addDependency call between fetchOperation and runtimeFetchOperation

fearless/Modules/Staking/Operations/ValidatorOperationFactory.swift
```diff
@@ -172,6 +176,49 @@ final class ValidatorOperationFactory {
...
+            storageRequestFactory.queryItems(
+                engine: engine,
+                keyParams: { accountIdList },
+                factory: { try runtimeFetchOperation.extractNoCancellableResultData() },
+                storagePath: .validatorPrefs
+            )
+
+        fetchOperation.allOperations.forEach { $0.addDependency(runtimeFetchOperation) }
```
stas-litvinov, line 193: @ERussel dependencies are set here

fearless/Modules/Staking/SelectValidatorsFlow/SelectValidatorsStart/RecommendationsComposing.swift
```diff
@@ -1,8 +1,24 @@
...
+    var stakeReturn: Decimal { get }
+    var totalStake: Decimal { get }
+    var ownStake: Decimal { get }
+    var hasIdentity: Bool { get }
+    var oversubscribed: Bool { get }
+    var hasSlashes: Bool { get }
+    var blocked: Bool { get }
+}
+
 protocol RecommendationsComposing {
```
ERussel, line 14: What is the point of this complicated protocol. Firstly, I thought it should allow us to reuse some logic but still duplicating the clustering logic. Can we move clustering logic into the separate method of the protocol to reuse it in the implementations?

fearless/Modules/Staking/SelectValidatorsFlow/ValidatorSearch/ValidatorSearchViewController.swift
```diff
@@ -9,6 +9,12 @@ final class ValidatorSearchViewController: UIViewController, ViewHolder {
```
ERussel, line 9: I think we need to fix the input indicator of the search field. Currently it's blue, I think it should be white.

fearless/Modules/Staking/SelectValidatorsFlow/ValidatorSearch/ValidatorSearchInteractor.swift
```diff
@@ -1,7 +1,43 @@
...
+        self.validatorOperationFactory = validatorOperationFactory
+        self.operationManager = operationManager
+    }
+}
 
 extension ValidatorSearchInteractor: ValidatorSearchInteractorInputProtocol {
-    func setup() {
-        #warning("Not implemented")
+    func performValidatorSearch(accountId: AccountId) {
+        let operation = validatorOperationFactory
```
ERussel, line 21: What if we already searching in that case race condition can happen. I think we should store pending wrapper locally and cancel it in case new one comes.

fearless/Modules/Staking/SelectValidatorsFlow/ValidatorSearch/ValidatorSearchInteractor.swift
```diff
@@ -7,26 +7,34 @@ final class ValidatorSearchInteractor {
...
 extension ValidatorSearchInteractor: ValidatorSearchInteractorInputProtocol {
     func performValidatorSearch(accountId: AccountId) {
-        let operation = validatorOperationFactory
+        cancelSearch()
+
+        currentOperation = validatorOperationFactory
             .wannabeValidatorsOperation(for: [accountId])
 
-        operation.targetOperation.completionBlock = { [weak self] in
+        currentOperation!.targetOperation.completionBlock = { [weak self] in
```
ERussel, line 33: I think we can just unwrap optional

fearless/Modules/Staking/SelectValidatorsFlow/ValidatorSearch/ValidatorSearchInteractor.swift
```diff
@@ -7,26 +7,34 @@ final class ValidatorSearchInteractor {
...
             DispatchQueue.main.async {
                 do {
-                    let result = try operation.targetOperation.extractNoCancellableResultData()
-
-                    guard let validatorInfo = result.first else {
+                    guard let result = try self?.currentOperation?.targetOperation.extractNoCancellableResultData(),
```
ERussel, line 36: should we also nullify the currentOperation here?

stas-litvinov: To my understanding it is not really necessary, since the operation itself is not going to produce any data at this point and for cancel function it doesn't really matter whether currentOperation is will or not. However, it might be worth to make code more precise. Let's do it

ERussel: There are several points why it is good to nullify:

1) Operation can contain logic that retains other objects, so it is expected that the operation is released when it is finished

2) Operation occupies memory

The first point is the most important I think.

ERussel (approved):

PR #317, opened 2021-07-15
FLW-987 iOS. Can't pay many rewards(block limit).

stas-litvinov:
Fixed erroneous behavior:
* Calculate block limit
* Split payouts into blocks
* Estimate fee for each block
* Send each blocks separately

Updated tests


fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -13,10 +13,24 @@ protocol ExtrinsicOperationFactoryProtocol {
 
     func submit(
         _ closure: @escaping ExtrinsicBuilderClosure,
-        signer: SigningWrapperProtocol
+        signer: SigningWrapperProtocol,
+        nonceShift: UInt32
```
ERussel, line 17: The drawback of this approach is that nonce still requested for each extrinsic. Moreover it might be a problem because there is no guarantee that it will not change somewhere in the middle while a list of transaction is being submitted one by one. So I suggest to have a method that allows to submit multiple extrinsic at once. This allows to request nonce ones and increment it for each extrinsic . The simplest solution that comes to mind is to have a parameter numberOfExtrinsics. Then closure parameter type should be also changed and include index of extrinsic for which builder should be setup.

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,236 +86,277 @@ final class StakingPayoutConfirmationInteractor {
...
 
         operationManager.enqueue(operations: [queryOperation], in: .transient)
     }
+}
+
+// MARK: - RuntimeConstantFetching
+
+extension StakingPayoutConfirmationInteractor: RuntimeConstantFetching {
+    private func createBatches() {
+        fetchCompoundConstant(
```
ERussel, line 233: In this case use operations chaining it is faster and simpler than closures.

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,236 +86,277 @@ final class StakingPayoutConfirmationInteractor {
...
-            let stashItem = changes.reduceToLastChange()
-            self?.handle(stashItem: stashItem)
-        }
+// MARK: - StakingPayoutConfirmationInteractorInputProtocol
 
-        let failureClosure: (Error) -> Void = { [weak self] error in
-            self?.presenter.didReceive(stashItemError: error)
+extension StakingPayoutConfirmationInteractor: StakingPayoutConfirmationInteractorInputProtocol {
+    func setup() {
+        guard let selectedAccountAddress = settings.selectedAccount?.address else {
```
ERussel, line 130: Is there any use case for settings other then fetching selected account? It might be simpler to pass the account address directly to interactor

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,236 +86,277 @@ final class StakingPayoutConfirmationInteractor {
...
-                self?.presenter.didReceive(feeError: error)
+extension StakingPayoutConfirmationInteractor: ExtrinsicFeeProxyDelegate {
+    enum FeeType: String {
+        case elaborateBlockWeight
+        case estimateFirstBlock
+        case estimateLastBlock
+    }
+
+    func didReceiveFee(result: Result<RuntimeDispatchInfo, Error>, for identifier: ExtrinsicFeeId) {
+        switch result {
```
ERussel, line 305: I think it it more simpler and clear to use ExtrinsicOperationFactory directly for fee calculation because one can benifit operation chaining. You can even chain it with constant extraction from runtime.

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationPresenter.swift
```diff
@@ -139,96 +89,167 @@ extension StakingPayoutConfirmationPresenter: StakingPayoutConfirmationPresenter
...
-extension StakingPayoutConfirmationPresenter: StakingPayoutConfirmationInteractorOutputProtocol {
-    func didReceive(stashItem: StashItem?) {
-        if let stashItem = stashItem {
-            logger?.debug("Stash: \(stashItem.stash)")
-            logger?.debug("Controller: \(stashItem.controller)")
-        } else {
-            logger?.debug("No stash found")
-        }
+    func proceed() {
+        guard let fee = fee else {
```
ERussel, line 99: We might want to integrate data validator here

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationPresenter.swift
```diff
@@ -139,96 +89,167 @@ extension StakingPayoutConfirmationPresenter: StakingPayoutConfirmationPresenter
...
+                balance = 0.0
+            }
+
+        case let .failure(error):
+            logger?.error("Account Info subscription error: \(error)")
         }
+    }
 
-        provideFee()
+    func didReceiveStashItem(result: Result<StashItem?, Error>) {
```
ERussel, line 183: We just don't need stash in presenter other than to create reward destination. I think it is simpler to store it in interactor as we did for reward destination change module

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationPresenter.swift
```diff
@@ -139,96 +89,167 @@ extension StakingPayoutConfirmationPresenter: StakingPayoutConfirmationPresenter
...
 
-    func didReceive(balance: AccountData?) {
-        if let availableValue = balance?.available {
-            self.balance = Decimal.fromSubstrateAmount(
-                availableValue,
-                precision: asset.precision
-            )
-        } else {
-            self.balance = 0.0
+    func didReceiveRewardDestiination(result: Result<RewardDestination<DisplayAddress>?, Error>) {
```
ERussel, line 225: Double ii.

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationPresenter.swift
```diff
@@ -139,96 +89,167 @@ extension StakingPayoutConfirmationPresenter: StakingPayoutConfirmationPresenter
...
+
+        wireframe.complete(from: view)
     }
 
-    func didReceive(priceError: Error) {
-        handle(error: priceError)
+    func didFailPayout(error: Error) {
+        view?.didStopLoading()
+
+        handle(error: error)
```
ERussel, line 252: I think we don't show an error in handle method if it is unknown. Please, show CommonError.undefined instead of just logging

fearlessTests/Mocks/DataProviders/ExtrinsicOperationFactoryStub.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    func estimateFeeOperation(_ closure: @escaping ExtrinsicBuilderIndexedClosure, numberOfExtrinsics: Int) -> CompoundOperationWrapper<[FeeExtrinsicResult]> {
+        let dispatchInfo = RuntimeDispatchInfo(dispatchClass: "Extrinsic",
+                                               fee: "10000000000",
+                                               weight: 10005000)
+
+        return CompoundOperationWrapper.createWithResult([.success(dispatchInfo)])
+    }
+
+    func submit(_ closure: @escaping ExtrinsicBuilderClosure, signer: SigningWrapperProtocol) -> CompoundOperationWrapper<String> {
```
ERussel, line 19: Last two is redundant I guess

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,237 +105,295 @@ final class StakingPayoutConfirmationInteractor {
...
         }
+    }
 
-        let failureClosure = { [weak self] (error: Error) in
-            self?.presenter.didReceive(priceError: error)
-            return
+    private func createFeeOperationWrapper() -> CompoundOperationWrapper<Decimal>? {
+        guard let batches = batches else { return nil }
+
+        let feeBatches: [Batch] = batches.enumerated().compactMap {
```
ERussel, line 167: Why it is so complicated? Please, do something like `let fetchBatches = batches.count > 1 ? [batches[0], batches[count-1]] : [batches[0]]

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,237 +105,295 @@ final class StakingPayoutConfirmationInteractor {
...
 
-        let failureClosure: (Error) -> Void = { [weak self] error in
-            self?.presenter.didReceive(stashItemError: error)
-            return
+                case let .failure(error):
+                    throw error
+                }
+            }
+
+            return (fees.first ?? 0.0) + (fees.last ?? 0.0) * Decimal(batches.count - 1)
```
ERussel, line 197: Looks like we need to do `first * (count - 1) + last`

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,237 +105,295 @@ final class StakingPayoutConfirmationInteractor {
...
+        } catch {
+            logger?.error("Did receive reward destination error: \(error)")
         }
+    }
 
-        let failureClosure = { [weak self] (error: Error) in
-            self?.presenter.didReceive(priceError: error)
-            return
+    private func createFeeOperationWrapper() -> CompoundOperationWrapper<Decimal>? {
+        guard let batches = batches else { return nil }
```
ERussel, line 165: And probably not empty

fearless/Modules/Staking/StakingPayoutConfirmation/StakingPayoutConfirmationInteractor.swift
```diff
@@ -84,237 +105,295 @@ final class StakingPayoutConfirmationInteractor {
...
+            runningIn: .main,
+            numberOfExtrinsics: batches.count
         ) { [weak self] result in
             switch result {
-            case let .success(txHash):
-                self?.presenter.didCompletePayout(txHash: txHash)
+            case let .success(extrinsicResultList):
+                do {
+                    let txHashes: [String] = try extrinsicResultList.map { result in
+                        switch result {
```
ERussel, line 309: `try result.get()` ?

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -24,12 +94,27 @@ protocol ExtrinsicServiceProtocol {
         completion completionClosure: @escaping EstimateFeeClosure
```
ERussel, line 94: Please, remove. We don't need this anymore

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -24,12 +94,27 @@ protocol ExtrinsicServiceProtocol {
...
+        runningIn queue: DispatchQueue,
+        numberOfExtrinsics: Int,
+        completion completionClosure: @escaping EstimateFeeIndexedClosure
+    )
+
     func submit(
         _ closure: @escaping ExtrinsicBuilderClosure,
         signer: SigningWrapperProtocol,
         runningIn queue: DispatchQueue,
         completion completionClosure: @escaping ExtrinsicSubmitClosure
```
ERussel, line 108: Please, remove. We don't need this anymore

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -122,38 +215,60 @@ extension ExtrinsicOperationFactory: ExtrinsicOperationFactoryProtocol {
...
-        infoOperation.addDependency(builderOperation)
+        feeOperationList.forEach { feeOperation in
+            wrapperOperation.addDependency(feeOperation)
+        }
 
         return CompoundOperationWrapper(
-            targetOperation: infoOperation,
-            dependencies: [nonceOperation, codingFactoryOperation, builderOperation]
+            targetOperation: wrapperOperation,
+            dependencies: [nonceOperation, codingFactoryOperation, builderOperation] + wrapperOperation.dependencies
```
ERussel, line 263: Please, put manually `feeOperationList` otherwise we can at some point add an operation twice

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -165,33 +280,54 @@ extension ExtrinsicOperationFactory: ExtrinsicOperationFactoryProtocol {
...
 
-        submitOperation.addDependency(builderOperation)
+        submitOperationList.forEach { submitOperation in
+            wrapperOperation.addDependency(submitOperation)
+        }
 
         return CompoundOperationWrapper(
-            targetOperation: submitOperation,
-            dependencies: [nonceOperation, codingFactoryOperation, builderOperation]
+            targetOperation: wrapperOperation,
```
ERussel, line 329: Please, put manually feeOperationList otherwise we can at some point add an operation twice

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -165,33 +280,54 @@ extension ExtrinsicOperationFactory: ExtrinsicOperationFactoryProtocol {
...
+            dependencies: [nonceOperation, codingFactoryOperation, builderOperation] + wrapperOperation.dependencies
         )
```
ERussel, line 331: Please, put manually feeOperationList otherwise we can at some point add an operation twice

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -4,17 +4,87 @@ import RobinHood
 import IrohaCrypto
 
 typealias ExtrinsicBuilderClosure = (ExtrinsicBuilderProtocol) throws -> (ExtrinsicBuilderProtocol)
-typealias EstimateFeeClosure = (Result<RuntimeDispatchInfo, Error>) -> Void
-typealias ExtrinsicSubmitClosure = (Result<String, Error>) -> Void
+typealias ExtrinsicBuilderIndexedClosure = (ExtrinsicBuilderProtocol, Int) throws -> (ExtrinsicBuilderProtocol)
+
+typealias FeeExtrinsicResult = Result<RuntimeDispatchInfo, Error>
+typealias EstimateFeeClosure = (FeeExtrinsicResult) -> Void
+typealias EstimateFeeIndexedClosure = (Result<[FeeExtrinsicResult], Error>) -> Void
```
ERussel, line 11: I think this look weird. I suggest to assume that if something failed before extrinsics are sent then assume that all extrinsics are failed and repeat error for all extrinsic results. Then the closure will be just `[Result<...>] -> Void`

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
stas-litvinov: Do you suggest to create wrappers instead?

ERussel: We have this in extension, so we don't need to request implementation of it in protocol

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -388,10 +388,15 @@ extension ExtrinsicService: ExtrinsicServiceProtocol {
...
+                if let operationResult = wrapper.targetOperation.result {
+                    do {
+                        let results = try operationResult.get()
+                        completionClosure(results)
+                    } catch {
+                        completionClosure([.failure(error)])
+                    }
                 } else {
-                    completionClosure(.failure(BaseOperationError.parentOperationCancelled))
+                    completionClosure([.failure(BaseOperationError.parentOperationCancelled)])
```
ERussel, line 399: I think we should be consistent with the number of extrinsics here

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -431,10 +436,15 @@ extension ExtrinsicService: ExtrinsicServiceProtocol {
...
```
ERussel, line 447: I think we should be consistent with the number of extrinsics here

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -259,4 +424,41 @@ extension ExtrinsicService: ExtrinsicServiceProtocol {
...
+        signer: SigningWrapperProtocol,
+        runningIn queue: DispatchQueue,
+        numberOfExtrinsics: Int,
+        completion completionClosure: @escaping ExtrinsicSubmitIndexedClosure
+    ) {
+        let wrapper = operationFactory.submit(closure, signer: signer, numberOfExtrinsics: numberOfExtrinsics)
+
+        wrapper.targetOperation.completionBlock = {
+            queue.async {
+                if let operationResult = wrapper.targetOperation.result {
```
ERussel, line 439: Please, use extractNonCancellable result to avoid duplicated code for error handling

fearless/Common/Services/ExtrinsicService/ExtrinsicService.swift
```diff
@@ -239,6 +375,35 @@ extension ExtrinsicService: ExtrinsicServiceProtocol {
...
+        completion completionClosure: @escaping EstimateFeeIndexedClosure
+    ) {
+        let wrapper = operationFactory.estimateFeeOperation(
+            closure,
+            numberOfExtrinsics: numberOfExtrinsics
+        )
+
+        wrapper.targetOperation.completionBlock = {
+            queue.async {
+                if let operationResult = wrapper.targetOperation.result {
```
ERussel, line 391: Please, use extractNonCancellable result to avoid duplication code for error handling

ERussel (approved):

PR #318, opened 2021-07-19
FLW-1031 Calculate "estimating era completion time"

ef1rspb:


fearless/Common/Services/EraCountdownService/EraCountdownService.swift
```diff
@@ -0,0 +1,203 @@
...
+
+protocol EraCountdownServiceProtocol {
+    func fetchCountdownOperationWrapper() -> CompoundOperationWrapper<UInt64>
+}
+
+enum EraCountdownServiceError: Error {
+    case noData
+}
+
+final class EraCountdownService: EraCountdownServiceProtocol {
```
ERussel, line 13: Let't rename this class to EraCountdownOperationFactory for consistency because it just creates operations and doesn't perform any other logic

fearless/Common/Services/EraCountdownService/EraCountdownService.swift
```diff
@@ -0,0 +1,203 @@
...
+    let chain: Chain
```
ERussel, line 14: Don't see how we are using this field

fearless/Common/Services/EraCountdownService/EraCountdownService.swift
```diff
@@ -0,0 +1,203 @@
...
+        storageRequestFactory: StorageRequestFactoryProtocol,
+        engine: JSONRPCEngine
+    ) {
+        self.chain = chain
+        self.runtimeCodingService = runtimeCodingService
+        self.storageRequestFactory = storageRequestFactory
+        self.engine = engine
+    }
+
+    func fetchCountdownOperationWrapper() -> CompoundOperationWrapper<UInt64> {
```
ERussel, line 31: In some cases, for example, for unbondings and payouts we need to display time from now till some era. In current solution we need first to fetch remained time from in current era and then calculate difference between eras. And so there are two problems: 1) we should do several steps 2) we use hardcoded assumption about era duration. My suggestion is to include here an optional closure parameter (so we could chain with other operations) that returns destination era.

PR #319, opened 2021-07-20
FLW-1030 Countdown timer: Unbondings, Waiting for next era, Pending rewards

ef1rspb:
https://user-images.githubusercontent.com/17689921/126506480-c4f02790-96ea-4853-ab92-58d89f635d32.mov


fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
```diff
@@ -0,0 +1,29 @@
...
+        let numberOfSlotsPerSession = UInt64(sessionLength)
+        let currentSessionIndexInt = UInt64(currentSessionIndex)
+
+        let sessionStartSlot = currentSessionIndexInt * numberOfSlotsPerSession + genesisSlot
+        let sessionProgress = currentSlot - sessionStartSlot
+        let eraProgress = (currentSessionIndexInt - UInt64(eraStartSessionIndex)) * numberOfSlotsPerSession + sessionProgress
+        let eraRemained = UInt64(eraLength) * numberOfSlotsPerSession - eraProgress
+        let result = eraRemained * UInt64(blockCreationTime)
+
+        let distanceInDates = Date()
```
ERussel, line 23: I think `Date().timeIntervalSinceDate(createdDate)` looks simpler or is there any trick?

fearless/Common/Extension/FearlessUtils/StorageKeyFactory+Implicit.swift
```diff
@@ -50,6 +50,13 @@ extension StorageKeyFactoryProtocol {
         )
     }
 
+    func currentSessionIndex() throws -> Data {
+        try createStorageKey(
```
ERussel, line 54: I think you can create StorageKey from StorageCodingPath for consistency

fearless/Common/Substrate/Types/Typealiases.swift
```diff
@@ -0,0 +1,3 @@
+typealias Slot = UInt64
```
ERussel, line 1: I think we already have `Common/Model/SubstrateAliases.swift`

fearless/Modules/Staking/StakingBalance/ViewModel/StakingBalanceViewModelFactory.swift
```diff
@@ -130,24 +132,17 @@ struct StakingBalanceViewModelFactory: StakingBalanceViewModelFactoryProtocol {
...
-                    .fromSubstrateAmount(
-                        unbondingItem.value,
-                        precision: precision
-                    ) ?? .zero
-                let tokenAmount = tokenAmountText(unbondingAmountDecimal, locale: locale)
-                let usdAmount = priceText(unbondingAmountDecimal, priceData: priceData, locale: locale)
-                let daysLeft = daysLeftAttributedString(
-                    activeEra: balanceData.activeEra,
-                    unbondingEra: unbondingItem.era,
+        (0 ..< 4)
```
ERussel, line 138: Looks like incomplete implemetation

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    var eraCompletionTime: TimeInterval {
+        let numberOfSlotsPerSession = UInt64(sessionLength)
+        let currentSessionIndexInt = UInt64(currentSessionIndex)
+
+        let sessionStartSlot = currentSessionIndexInt * numberOfSlotsPerSession + genesisSlot
+        let sessionProgress = currentSlot - sessionStartSlot
+        let eraProgress = (currentSessionIndexInt - UInt64(eraStartSessionIndex)) * numberOfSlotsPerSession
+            + sessionProgress
+        let eraRemained = UInt64(eraLength) * numberOfSlotsPerSession - eraProgress
```
ERussel, line 21: We might run into the situation when currentIndex already past current era but due to network lag we still on the old era. Let's put a check here before subtracting

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
```diff
@@ -0,0 +1,28 @@
...
+        let result = eraRemained * UInt64(blockCreationTime)
+
+        let datesTimeinterval = Date().timeIntervalSince(createdAtDate)
+        let remainedTime = TimeInterval(result) - datesTimeinterval
```
ERussel, line 25: I think result here must be converted to seconds because `datesTimeinterval` is in seconds

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdownOperationFactory.swift
```diff
@@ -0,0 +1,141 @@
...
+        storageRequestFactory: StorageRequestFactoryProtocol,
+        engine: JSONRPCEngine
+    ) {
+        self.runtimeCodingService = runtimeCodingService
+        self.storageRequestFactory = storageRequestFactory
+        self.engine = engine
+    }
+
+    // swiftlint:disable function_body_length
+    func fetchCountdownOperationWrapper(targetEra: EraIndex) -> CompoundOperationWrapper<EraCountdown> {
```
ERussel, line 29: If we don't want to calculate time between eras here (we are doing here somewhere in view model factories) then this parameter is useless and error prone. Let's just get rid of it and fetch active era manually

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+struct EraCountdown {
```
ERussel, line 3: Let's also add active era var

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
ef1rspb: Now we manually fetch activeEra, so the these situation is impossible, right?

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdownOperationFactory.swift
ef1rspb: done!

fearless/Common/Operation/EraCountdownOperationFactory/EraCountdown.swift
ERussel: I think yes. Because we don't fetch active era and currentSessionIndex atomically. So, it is better to add comparison and in case of negative result return 0

ERussel (approved):

PR #321, opened 2021-07-21
FLW-1033 Waiting for next era: countdown

ef1rspb:
https://user-images.githubusercontent.com/17689921/126515566-02a5924c-3635-4e19-a252-700c79444422.mov


fearless/Modules/Staking/StakingMain/StakingMainPresenter.swift
```diff
@@ -510,6 +510,16 @@ extension StakingMainPresenter: StakingMainInteractorOutputProtocol {
         }
     }
 
+    func didReceive(eraCompletionTimeResult: Result<UInt64, Error>) {
+        switch eraCompletionTimeResult {
+        case let .success(time):
+            let eraCompletionTimeInSeconds = TimeInterval(time / 1000)
```
ERussel, line 516: We have TimeInterval extension that contains conversions (milliseconds, seconds, dayFromSeconds)

fearless/Modules/Staking/StakingMain/StakingMainInteractor+InputProtocol.swift
```diff
@@ -18,6 +18,7 @@ extension StakingMainInteractor: StakingMainInteractorInputProtocol {
         provideRewardCalculator()
         provideEraStakersInfo()
         provideNetworkStakingInfo()
+        fetchEraCompletionTime()
```
ERussel, line 21: We should find home for this call. 

The problem here is that we could receive this info in some time in the past and at some point in future the state of NominatorView changes and we are starting displaying incorrect timer. Probably we should store local time when the value was fetched and then we can always correct this value by providing current local time. 

Also we should request new value when era changes. I think we can pair this request with `provideEraStakersInfo` call because the interactor is notified when this happens and basically this means that era changed.

ERussel (approved):

PR #329, opened 2021-07-26
FLW-1074 Enter pin screen: add Face ID icon and action

ef1rspb:


ERussel (approved):

fearless/Modules/Pincode/PinSetup/PinSetupViewController.swift
```diff
@@ -245,8 +245,9 @@ extension PinSetupViewController: PinSetupViewProtocol {
         }
     }
 
-    func didChangeAccessoryState(enabled: Bool) {
+    func didChangeAccessoryState(enabled: Bool, accessoryIcon: UIImage?) {
```
ERussel, line 248: Maybe pass type here instead of an icon so we could check in tests that we detected all correctly

ERussel (approved):

PR #363, opened 2021-08-13
Runtime sync

ERussel:
Add runtime sync service that syncs types and metadata for each chain and controls number of operations executed concurrently to prevent out of memory problems

### DIFF
--- a/fearless/Common/EventCenter/EventVisitor.swift
+++ b/fearless/Common/EventCenter/EventVisitor.swift
@@ -14,6 +14,10 @@ protocol EventVisitorProtocol: AnyObject {
     func processChainSyncDidStart(event: ChainSyncDidStart)
     func processChainSyncDidComplete(event: ChainSyncDidComplete)
     func processChainSyncDidFail(event: ChainSyncDidFail)
+
+    func processRuntimeBaseTypesSyncCompleted(event: RuntimeBaseTypesSyncCompleted)
+    func processRuntimeChainTypesSyncCompleted(event: RuntimeChainTypesSyncCompleted)
+    func processRuntimeChainMetadataSyncCompleted(event: RuntimeMetadataSyncCompleted)
 }
 
 extension EventVisitorProtocol {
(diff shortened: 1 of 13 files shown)


fearless/Common/EventCenter/EventVisitor.swift
```diff
@@ -14,6 +14,10 @@ protocol EventVisitorProtocol: AnyObject {
     func processChainSyncDidStart(event: ChainSyncDidStart)
     func processChainSyncDidComplete(event: ChainSyncDidComplete)
     func processChainSyncDidFail(event: ChainSyncDidFail)
+
+    func processRuntimeBaseTypesSyncCompleted(event: RuntimeBaseTypesSyncCompleted)
```
ef1rspb, line 18: I guess EventVisitor interface contains too many methods, can we loot to interface segregation?

fearless/Common/Model/ChainRegistry/ChainModelMapper.swift
```diff
@@ -38,20 +38,30 @@ extension ChainModelMapper: CoreDataMapperProtocol {
...
             icon: entity.icon!,
             isEthereumBased: entity.isEthereumBased
         )
     }
 
     func populate(entity: CDChain, from model: ChainModel, using context: NSManagedObjectContext) throws {
         entity.chainId = model.chainId
-        entity.types = model.types
+        entity.types = model.types?.url
+        entity.typesOverrideCommon = model.types.map { NSNumber(value: $0.overridesCommon) }
```
ef1rspb, line 63: maybe `NSNumber(booleanLiteral: )`?

fearless/Common/EventCenter/EventVisitor.swift
ERussel: So the purpose of the EventCenter is to be the mediator for event delivery and visitor pattern helps to deal with many different types without switch...cases. But I will think how we could improve this if it grows abnormal.

PR #365, opened 2021-08-18
FLW-650 Update background and Splash screen

stas-litvinov:
- Change bottom inset for buttons to 16px from Safe area
- Change background color for action sheets to opaque black
- Change disabled triangulated button appearance

### DIFF
--- a/fearless/Common/Extension/UIKit/TriangularedButton+Style.swift
+++ b/fearless/Common/Extension/UIKit/TriangularedButton+Style.swift
@@ -2,6 +2,16 @@ import Foundation
 import SoraUI
 
 extension TriangularedButton {
+    override var isEnabled: Bool {
+        didSet {
+            if isEnabled {
+                applyEnabledStyle()
+            } else {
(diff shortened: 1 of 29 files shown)


fearless/Common/Extension/UIKit/TriangularedButton+Style.swift
```diff
@@ -2,6 +2,16 @@ import Foundation
 import SoraUI
 
 extension TriangularedButton {
+    override var isEnabled: Bool {
+        didSet {
```
ERussel, line 6: I think we can have problems with this. Swift docs also doesn't like this: Extensions can add new computed properties, but they can’t add stored properties, or add property observers to existing properties. https://docs.swift.org/swift-book/LanguageGuide/Extensions.html#//apple_ref/doc/uid/TP40014097-CH24-ID151

stas-litvinov: Let's try another way

ERussel (approved):

PR #372, opened 2021-08-27
FLW-1220: iOS: Remove limitation of staking parameters (e.g. Validators) when maximum validators are reached

ERussel:
- remove restriction on max number of nominators on validators change

### DIFF
--- a/fearless/Modules/Staking/SelectValidatorsFlow/SelectValidatorsConfirm/SelectValidatorsConfirmPresenter.swift
+++ b/fearless/Modules/Staking/SelectValidatorsFlow/SelectValidatorsConfirm/SelectValidatorsConfirmPresenter.swift
@@ -156,29 +156,47 @@ extension SelectValidatorsConfirmPresenter: SelectValidatorsConfirmPresenterProt
     }
 
     func proceed() {
+        guard let state = state else {
+            return
+        }
+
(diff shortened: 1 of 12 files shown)


fearless/Modules/Staking/SelectValidatorsFlow/SelectValidatorsConfirm/SelectValidatorsConfirmPresenter.swift
```diff
@@ -156,29 +156,47 @@ extension SelectValidatorsConfirmPresenter: SelectValidatorsConfirmPresenterProt
...
+                hasExistingNomination: state.hasExistingNomination,
                 locale: locale
             )
-        ]).runValidation { [weak self] in
-            guard let fee = self?.fee else {
-                return
-            }
+        ]
+
+        if !state.hasExistingBond {
```
ef1rspb, line 187: If there is no bond how we can validate that?

ERussel: If we don't have existing bonding then we are going to submit new one with validators and we need to apply validation for the bonding value

ef1rspb (approved):

PR #386, opened 2021-09-09
FLW-1270 iOS: Migrate from Subscan to Coingecko API for USD representation of KSM & DOT

stas-litvinov:
Replace Subscan API with Coingecko API


fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
```diff
@@ -0,0 +1,103 @@
...
+            )
+
+            request.httpMethod = HttpMethod.get.rawValue
+
+            return request
+        }
+
+        let resultFactory = AnyNetworkResultFactory<PriceData?> { data in
+            let priceData = try JSONDecoder().decode(
+                CoingeckoPriceData.self,
```
ERussel, line 92: Why do we need to deal with dynamic keys if we could just parse as `[String: PriceData].self`?

stas-litvinov: I tried, but it did not work. Later I found another problem with mapping and decided to leave it as it is now. I guess, we'll need it later

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
```diff
@@ -0,0 +1,103 @@
...
+    static let baseUrlString = "https://api.coingecko.com/api/v3"
+
+    private func buildURLForAssets(
+        _ assets: [WalletAssetId],
+        method: String,
+        currencies: [String] = ["usd"],
+        options: CoingeckoPriceRequestOptions = [.includeDayChange, .includeLastUpdatedAt]
+    ) -> URL? {
+        guard var components = URLComponents(
+            string: Self.baseUrlString + method
```
ERussel, line 36: It is more safe to use `appendingPathComponent`

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
```diff
@@ -0,0 +1,113 @@
...
+import RobinHood
+
+protocol CoingeckoOperationFactoryProtocol {
+    func fetchPriceOperation(for assets: [WalletAssetId]) -> BaseOperation<[PriceData]>
+}
+
+struct CoingeckoPriceRequestOptions: OptionSet {
+    let rawValue: Int
+
+    static let includeMarketCap = CoingeckoPriceRequestOptions(rawValue: 1 << 0)
```
antonkhvorov, line 11: We don't need it

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
```diff
@@ -0,0 +1,113 @@
...
+    static let includeDayVolume = CoingeckoPriceRequestOptions(rawValue: 1 << 1)
```
antonkhvorov, line 12: We don't need it

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
```diff
@@ -0,0 +1,113 @@
...
+    static let includeDayChange = CoingeckoPriceRequestOptions(rawValue: 1 << 2)
+    static let includeLastUpdatedAt = CoingeckoPriceRequestOptions(rawValue: 1 << 3)
```
antonkhvorov, line 14: We don't need it

antonkhvorov (changes requested):

antonkhvorov:
Let keep things nice & simple, especially to the parts which are going to be reworked in 2.0 (such as this feature).
This PR should be straightforward & small, so that there will be no room for errors since we need to release it asap

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
ERussel: fixed

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
ERussel: fixed

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
ERussel: fixed

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
ERussel: fixed

fearless/Common/Network/Coingecko/CoingeckoOperationFactory.swift
ERussel: fixed

antonkhvorov (approved):

PR #388, opened 2021-09-14
FLW-1273 Account mappers

ERussel:
- add account mapper for new account entity
- add compatable mapper for legacy AccountItem
- add shared account settings to store selected wallet

Note: it is ok that tests are failing now, it will be fixed in next PRs


fearless/Modules/Staking/StakingMain/StakingMainViewFactory.swift
```diff
@@ -79,15 +79,15 @@ final class StakingMainViewFactory: StakingMainViewFactoryProtocol {
             runtimeService: runtimeService
         )
 
-        let repository: CoreDataRepository<AccountItem, CDAccountItem> =
-            UserDataStorageFacade.shared.createRepository()
-
-        let accountRepositoryFactory = AccountRepositoryFactory(
+        // TODO: check why do we need both factory and repository
```
stas-litvinov, line 82: It builds well without `accountRepository`, so it looks like we don't need it

fearless/Common/Storage/EntityToModel/ManagedMetaAccountMapper.swift
```diff
@@ -0,0 +1,56 @@
...
+            let sortDescriptor = NSSortDescriptor(key: #keyPath(CDMetaAccount.order), ascending: false)
+            fetchRequest.sortDescriptors = [sortDescriptor]
+            fetchRequest.predicate = NSPredicate(format: "%K > 0", #keyPath(CDMetaAccount.order))
+            fetchRequest.fetchLimit = 1
+
+            let maybeLastItem = try context.fetch(fetchRequest).first
+
+            order = maybeLastItem?.order ?? 0
+
+            entity.order = order + 1
```
stas-litvinov, line 51: So in this case there will be no metas with `.oder == 0`?

fearless/Common/Storage/PersistentValueSettings.swift
```diff
@@ -0,0 +1,84 @@
...
+            mutex.unlock()
+        }
+
+        return internalValue
+    }
+
+    var hasValue: Bool { value != nil }
+
+    func performSetup(completionClosure _: @escaping (Result<T?, Error>) -> Void) {
+        // Function must be implemented in subclass
```
stas-litvinov, line 27: Isn't it safer to throw an error here?

fearless/Modules/Root/RootInteractor.swift
```diff
@@ -78,5 +78,19 @@ extension RootInteractor: RootInteractorInputProtocol {
...
+        settings.setup(runningCompletionIn: .main) { result in
+            switch result {
+            case let .success(maybeMetaAccount):
+                if let metaAccount = maybeMetaAccount {
+                    self.logger?.debug("Selected account: \(metaAccount.metaId)")
+                } else {
+                    self.logger?.debug("No selected account")
+                }
+            case let .failure(error):
+                self.logger?.error("Selected account setup failed: \(error)")
```
stas-litvinov, line 92: Do we only log the result here? Are there any other actions?

stas-litvinov (approved):

fearless/Modules/Staking/StakingMain/StakingMainViewFactory.swift
ERussel: Ok, we will try to get rid of it when we integrate shared state

fearless/Common/Storage/EntityToModel/ManagedMetaAccountMapper.swift
ERussel: Right. We have 0 to mean that no order was set for the item.

fearless/Common/Storage/PersistentValueSettings.swift
ERussel: Yep, we could add `fatalError("Must be implemented on inheritance")`. Will do in the next PR.

fearless/Modules/Root/RootInteractor.swift
ERussel: I think we could add error state but this case can happen when something broken in the database. As we assume that a user will not change db manually then it is ok for now just log the error and show onboarding screen.

PR #389, opened 2021-09-15
Crowdloan & Staking settings

ERussel:
- add settings to store selected chain for crowloan
- add settings to store selected chain and asset for staking


fearless/Common/Storage/CrowdloanChainSettings.swift
```diff
@@ -0,0 +1,81 @@
...
+
+            repository = AnyDataProviderRepository(
+                storageFacade.createRepository(filter: filter, sortDescriptors: [], mapper: mapper)
+            )
+        } else {
+            let filter = NSPredicate.relayChains()
+            repository = AnyDataProviderRepository(
+                storageFacade.createRepository(filter: filter, sortDescriptors: [], mapper: mapper)
+            )
+        }
```
stas-litvinov, line 40: I would suggest to get rid of duplicated code:
```swift
let filter: NSPredicate = {
    if let chainId = maybeChainId {
        return NSCompoundPredicate(orPredicateWithSubpredicates: [
            NSPredicate.chainBy(identifier: chainId),
            NSPredicate.relayChains()
        ])
    } else {
        return NSPredicate.relayChains()
    }
}()

repository = AnyDataProviderRepository(
    storageFacade.createRepository(filter: filter, sortDescriptors: [], mapper: mapper)
)
```

fearless/Common/Storage/StakingAssetSettings.swift
```diff
@@ -0,0 +1,93 @@
...
+                let firstRelayChain = relayChains.min(by: { $0.addressPrefix < $1.addressPrefix }),
+                let asset = firstRelayChain.assets.first(where: { $0.isUtility }) {
+                self.settings.stakingAsset = ChainAssetId(
+                    chainId: firstRelayChain.chainId,
+                    assetId: asset.assetId
+                )
+                return ChainAsset(chain: firstRelayChain, asset: asset)
+            }
+
+            return nil
```
stas-litvinov, line 65: Looks like this `if` block could be replaced with 
```swift
guard let ...
else { return nil }

self.settings.stakingAsset = ChainAssetId(
    chainId: firstRelayChain.chainId,
    assetId: asset.assetId
)
return ChainAsset(chain: firstRelayChain, asset: asset)
```

stas-litvinov (approved):

fearless/Common/Storage/StakingAssetSettings.swift
ERussel: This block of code consists from sequence of `if` statements. I think if we use guard instead of second `if` then it will become less readable

PR #399, opened 2021-09-27
Adopt account management to new model

ERussel:


fearless/Modules/AccountManagement/AccountManagementPresenter.swift
```diff
@@ -8,56 +8,31 @@ final class AccountManagementPresenter {
...
-
     let viewModelFactory: ManagedAccountViewModelFactoryProtocol
-    let supportedNetworks: [SNAddressType]
 
-    private var sections: [ManagedAccountViewModelSection] = []
+    private var viewModels: [ManagedAccountViewModelItem] = []
 
-    private let listCalculator: ListDifferenceCalculator<ManagedAccountItem> = {
-        let calculator = ListDifferenceCalculator<ManagedAccountItem>(initialItems: []) { item1, item2 in
+    private let listCalculator: ListDifferenceCalculator<ManagedMetaAccountModel> = {
```
stas-litvinov, line 15: I would call it `PositionCalculator` or `ListPositionCalculator`

fearless/Modules/AccountManagement/AccountManagementPresenter.swift
```diff
@@ -68,54 +43,53 @@ extension AccountManagementPresenter: AccountManagementPresenterProtocol {
...
+    func moveItem(at startIndex: Int, to finalIndex: Int) {
         guard startIndex != finalIndex else {
             return
         }
 
-        var newItems = sections[section].items
+        var newItems = viewModels
 
-        var saveItems: [ManagedAccountItem]
+        var saveItems: [ManagedMetaAccountModel]
```
stas-litvinov, line 84: Here I realised that new accounts order saves independently of whether user chooses to tap "Done" or "Back" buttons. Looks like it could be an unexpected behavior

fearless/Modules/AccountManagement/AccountManagementViewController.swift
```diff
@@ -78,10 +76,8 @@ final class AccountManagementViewController: UIViewController {
         tableView.contentInset = .init(top: 0, left: 0, bottom: bottomInset, right: 0)
 
         tableView.register(R.nib.accountTableViewCell)
-        tableView.register(
-            UINib(resource: R.nib.iconTitleHeaderView),
-            forHeaderFooterViewReuseIdentifier: Constants.headerId
-        )
+
+        tableView.rowHeight = Constants.cellHeight
```
stas-litvinov, line 80: Why do we need to set row height explicitly?

stas-litvinov (approved):

fearless/Modules/AccountManagement/AccountManagementViewController.swift
ERussel: Those cells doesn't support automatic height.

fearless/Modules/AccountManagement/AccountManagementPresenter.swift
ERussel: Need a pr to Robinhood, but basically the whole point of the class if to find changes

fearless/Modules/AccountManagement/AccountManagementPresenter.swift
ERussel: The same as for other actions like remove. I think this is normal behaviour in iOS. Edit button just shows controls for editing and Done - hides them

PR #407, opened 2021-10-05
FLW-1323 Fix Analytics crash: average amount is NaN

ef1rspb:


fearless/Common/View/Chart/FWBarChartView.swift
```diff
@@ -126,8 +126,15 @@ extension FWBarChartView: FWChartViewProtocol {
                 return chartAmount.filled ? chartAmount.value : 0.0
             }
         }
-        let (min, max) = (realAmounts.min() ?? 0.0, realAmounts.max() ?? 0.0)
-        averageLabelHeightPercent = (data.averageAmountValue - min) / (max - min)
+
+        if
+            let averageAmountValue = data.averageAmountValue,
+            let min = realAmounts.min(),
+            let max = realAmounts.max() {
```
stas-litvinov, line 133: If max == min then it will case division by 0 again, maybe it is worth to change it to be:
```swift
If
       let averageAmountValue = data.averageAmountValue,
       let min = realAmounts.min(),
       let max = realAmounts.max(),
       max != min
```

stas-litvinov: Or maybe
```swift
guard let min = realAmounts.min(),
      let max = realAmounts.max(),
      min != max else { ... } // throw an error?
...
```

fearless/Common/View/Chart/FWBarChartView.swift
```diff
@@ -126,8 +126,16 @@ extension FWBarChartView: FWChartViewProtocol {
...
-        averageLabelHeightPercent = (data.averageAmountValue - min) / (max - min)
+
+        if
+            let averageAmountValue = data.averageAmountValue,
+            let min = realAmounts.min(),
+            let max = realAmounts.max(),
+            max != min {
+            averageLabelHeightPercent = (averageAmountValue - min) / (max - min)
+        } else {
+            averageLabelHeightPercent = nil
```
ERussel, line 137: I think it should be 1 by L'Hôpital's rule

ef1rspb: Ok, so for max == min, we draw 100%, and for averageAmountValue == nil label is hidden

fearless/Modules/Staking/Analytics/Common/Module/AnalyticsViewModelFactoryBase.swift
```diff
@@ -112,15 +112,22 @@ class AnalyticsViewModelFactoryBase<T: AnalyticsViewModelItem> {
...
+            guard !chartDoubles.isEmpty else { return nil }
+            return chartDoubles.reduce(0.0, +) / Double(chartDoubles.count)
+        }()
+
+        let averageAmountText: String? = {
+            guard let averageAmount = averageAmount else { return nil }
+            let averageAmountRawText = balanceViewModelFactory
+                .amountFromValue(Decimal(averageAmount))
+                .value(for: locale)
+                .replacingOccurrences(of: " ", with: "\n")
```
ERussel, line 125: I suggest to use `AmountFormatterFactory.createDisplayFormater` to create amount without token symbol and the concate it with remained string that includes newline

ERussel (approved):

PR #408, opened 2021-10-05
FLW-1317 Analytics UI/UX improvements

ef1rspb:
- remove pinch-to-zoom from Rewards tab
- Analytics widget: KSM crosses month on OX (especially on small screen)
- Validators: doesn't fit into circle diagram, if validator doesn't have display name
- Validators: label"Staking was inactive" is smaller than "Staking was active”
- Staking tab: fix chart drops
- History: "Your rewards will appear hear" text is not relevant here, we need to change empty state to another text

### DIFF
--- a/fearless/Modules/Staking/Analytics/Stake/ViewModel/AnalyticsStakeViewModelFactory.swift
+++ b/fearless/Modules/Staking/Analytics/Stake/ViewModel/AnalyticsStakeViewModelFactory.swift
@@ -58,18 +58,96 @@ final class AnalyticsStakeViewModelFactory: AnalyticsViewModelFactoryBase<Subque
         by period: AnalyticsPeriod,
         locale: Locale
     ) -> [AnalyticsSelectedChartData] {
+        let dates = data.map(\.date)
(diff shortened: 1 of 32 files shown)


fearless/Modules/Staking/Analytics/Stake/ViewModel/AnalyticsStakeViewModelFactory.swift
```diff
@@ -27,18 +58,96 @@ final class AnalyticsStakeViewModelFactory: AnalyticsViewModelFactoryBase<Subque
...
+            let component: DateComponents = {
+                switch period {
+                case .month, .week:
+                    return DateComponents(day: index)
+                case .year, .all:
+                    return DateComponents(month: index)
+                }
+            }()
+            let currentDate = calendar.date(byAdding: component, to: timestampInterval.0) ?? startDate
+            let stakeChangesForCurrentDate = groupedByDate.map { date, stakeChanges -> (Date, [SubqueryStakeChangeData])? in
```
ERussel, line 103: Why not directly apply `compactMap` instead of `map.compactMap`?

ERussel (approved):

PR #457, opened 2021-11-18
Feature: (FLW-1371) subquery utils

Viscount88:


fearless/Common/Model/ChainModel.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import RobinHood
+
+struct ChainModel: Codable, Equatable {
+    var chainId: String
+    var name: String
+    var externalApi: ChainExternalApi?
+
+    static func == (lhs: ChainModel, rhs: ChainModel) -> Bool {
+        lhs.chainId == rhs.chainId && lhs.externalApi == rhs.externalApi
```
posplaw, line 10: why need to compare external api ? is it really possible to have multiple chain instances with different external apis?

fearless/Common/Model/ChainModel.swift
```diff
@@ -0,0 +1,30 @@
...
+    var history: ChainExternalApiObject
+
+    static func == (lhs: ChainExternalApi, rhs: ChainExternalApi) -> Bool {
+        lhs.staking == rhs.staking && lhs.history == rhs.history
+    }
+}
+
+struct ChainExternalApiObject: Codable {
+    var type: String
+    var url: String
```
posplaw, line 25: since we're on iOS, please if it's string, call it urlString, url can be named only if NSURL instance

fearless/Common/Repository/ChainLocalRepository.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+import RobinHood
+
+typealias ChainModelList = [ChainModel]
+
+extension ChainModelList: Identifiable {
+    public var identifier: String {
```
posplaw, line 7: why not let?

fearless/Common/Repository/ChainLocalRepository.swift
```diff
@@ -0,0 +1,44 @@
...
+        "ChainModelList"
+    }
+}
+
+final class ChainLocalRepository {
+    enum Constants {
+        static let githubChainListUrl: String = "https://raw.githubusercontent.com/soramitsu/fearless-utils/master/chains/chains.json"
```
posplaw, line 14: url string or make it url

fearless/Common/Repository/ChainLocalRepository.swift
```diff
@@ -0,0 +1,44 @@
...
+        )
+    }
+
+    func getSubqueryHistoryUrl(assetId: WalletAssetId) -> URL? {
+        let chains = utilsLocalRepository.fetch()
+
+        let urlString = chains?
+            .first { assetId.titleForLocale(Locale.current) == $0.name }?.externalApi?.history.url
+
+        guard let urlString = urlString, let url = URL(string: urlString) else {
```
posplaw, line 38: see, here you could do only one guard let to url itself

fearless/Common/Repository/UtilsLocalRepository.swift
```diff
@@ -0,0 +1,83 @@
...
+                logger?.error(error.localizedDescription)
+            }
+        }
+    }
+
+    func fetch() -> T? {
+        cache
+    }
+
+    private func readFromRepository() -> T? {
```
posplaw, line 39: so basically, if url was once loaded it's stuck in cache forever? with no overwrite even if changes are done?

bnsports: @posplaw no. every time you request url, it updates from remote repository and overwrite cached data.

fearless/Common/Repository/ChainLocalRepository.swift
Viscount88: Because we need extension to confirm protocol for typealias and swift don't allow extensions contain stored properties.

posplaw (approved):

PR #458, opened 2021-11-19
Fix/crowdloan validations

bnsports:


fearless/Common/Validation/Validators/BaseDataValidatorFactory.swift
```diff
@@ -119,7 +117,8 @@ extension BaseDataValidatingFactoryProtocol {
 
             if
                 let totalAmount = totalAmount,
-                let minimumBalance = minimumBalance {
+                let minimumBalance = minimumBalance,
+                totalAmount > spendingAmount {
```
Viscount88, line 121: minimumBalance can be negative number? Why we should check it additional to main check?

fearless/Modules/Crowdloan/CrowdloanContributionConfirm/CustomFlows/MoonbeamContributionConfirmInteractor.swift
```diff
@@ -109,14 +109,29 @@ class MoonbeamContributionConfirmInteractor: CrowdloanContributionConfirmInterac
             if let ethereumAccountAddress = self?.ethereumAccountAddress,
                let memoCall = self?.makeMemoCall(memo: ethereumAccountAddress) {
                 nextBuilder = try nextBuilder.adding(call: memoCall)
-
-                self?.saveEthereumAdressAsMoonbeamDefault()
```
Viscount88, line 113: Why we shouldn't save it here?

fearless/Common/Validation/Validators/BaseDataValidatorFactory.swift
bnsports: we are working with unsigned there so it cannot be negative. in case the result of (totalAmount - spendingAmount) will be negative - the app will be crashed. and because minimumBalance cannot be negative, we can say that 'if totalAmount < spendingAmount' that validation cannot be passed by it's logic.

fearless/Modules/Crowdloan/CrowdloanContributionConfirm/CustomFlows/MoonbeamContributionConfirmInteractor.swift
bnsports: it was a mistake to put it here. we have a such logic inside app: if user sends referral address once - don't put this address in future extrinsics because it fee-costable operation. by having save referral address here, we have a such case: if add_memo will be failed by some reason, app will not include this memo in future calls. so I replaced this saving to the place where we ensure that transaction was succeed

fearless/Common/Validation/Validators/BaseDataValidatorFactory.swift
Viscount88: got it

fearless/Common/Extension/ViewController+Wrapper.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit
+
+extension UIViewController {
+    func wrappedFromNavigationController() -> UIViewController? {
```
posplaw, line 4: Actually from the body of func I understand what's this, but name is weird

fearless/Modules/Crowdloan/CrowdloanContributionConfirm/CustomFlows/MoonbeamContributionConfirmInteractor.swift
```diff
@@ -109,14 +109,29 @@ class MoonbeamContributionConfirmInteractor: CrowdloanContributionConfirmInterac
...
+
+    override func submitContribution(builderClosure: @escaping ExtrinsicBuilderClosure) {
+        extrinsicService.submit(
+            builderClosure,
+            signer: signingWrapper,
+            runningIn: .main,
+            completion: { [weak self] result in
+                self?.confirmPresenter?.didSubmitContribution(result: result)
+
+                switch result {
```
posplaw, line 128: why not `if case .success = result` ? don't like such big constructions

Viscount88 (approved):

Viscount88 (approved):

posplaw (approved):

PR #474, opened 2021-12-21
Feature/(flw 1423) asset json handling

Viscount88:
Supporting separated assets json


Viscount88:
Another target branch

PR #480, opened 2021-12-28
Feature/flw 1409 buy flow

Viscount88:


fearless/Common/PurchaseProvider/MoonpayProvider.swift
```diff
@@ -26,22 +26,11 @@ final class MoonpayProvider: PurchaseProviderProtocol {
...
-        optionUrl = buildURLForToken("DOT", address: address)
-
-        if let url = optionUrl {
-            let action = PurchaseAction(title: "MoonPay", url: url, icon: R.image.iconMoonPay()!)
-            return [action]
-        } else {
-            return []
+    func buildPurchaseActions(asset: AssetModel, address: String) -> [PurchaseAction] {
+        if let url = buildURLForToken(asset.id, address: address) {
+            return [PurchaseAction(title: "Moonpay", url: url, icon: R.image.iconRamp()!)]
```
bnsports, line 31: string to constants

bnsports, line 31: why iconRamp in MoonpayProvider?

fearless/Common/PurchaseProvider/MoonpayProvider.swift
```diff
@@ -26,22 +26,11 @@ final class MoonpayProvider: PurchaseProviderProtocol {
...
-
-        guard chain == .polkadot else { return [] }
-        optionUrl = buildURLForToken("DOT", address: address)
-
-        if let url = optionUrl {
-            let action = PurchaseAction(title: "MoonPay", url: url, icon: R.image.iconMoonPay()!)
-            return [action]
-        } else {
-            return []
+    func buildPurchaseActions(asset: AssetModel, address: String) -> [PurchaseAction] {
```
bnsports, line 29: why array? is concrete provider can create multiple purchase actions?

fearless/Modules/NewWallet/ChainAccount/ChainAccountPresenter.swift
```diff
@@ -56,6 +66,25 @@ final class ChainAccountPresenter {
     }
 }
 
+private extension ChainAccountPresenter {
+    func getPurchaseActions() -> [PurchaseAction] {
+        var actions: [PurchaseAction] = []
+        if let address = selectedMetaAccount.fetch(for: chain.accountRequest())?.toAddress() {
+            let rampActions = rampProvider.buildPurchaseActions(
```
bnsports, line 73: [rampProvider, moonpayProvider].compactMap { $0.buildPurchaseActions(:) }.reduce([], +)

fearless/Modules/NewWallet/ChainAccount/ChainAccountPresenter.swift
```diff
@@ -115,3 +150,9 @@ extension ChainAccountPresenter: Localizable {
         provideViewModel()
     }
 }
+
+extension ChainAccountPresenter: ModalPickerViewControllerDelegate {
+    func modalPickerDidSelectModelAtIndex(_ index: Int, context _: AnyObject?) {
+        wireframe.presentPurchaseWebView(from: view, action: getPurchaseActions()[index])
```
bnsports, line 156: what is context here? maybe it will be safer to use it instead of getPurchaseActions(). in case order inside modal picker will be changed, current logic will not handle that case correctly

fearless/Modules/Purchase/PurchaseViewFactory.swift
```diff
@@ -5,15 +5,34 @@ import CommonWallet
 final class PurchaseViewFactory: PurchaseViewFactoryProtocol {
     static func createView(
         for action: PurchaseAction,
-        commandFactory: WalletCommandFactoryProtocol
+        commandFactory _: WalletCommandFactoryProtocol
```
bnsports, line 8: it looks like you can remove this parameter at all

fearless/Common/PurchaseProvider/MoonpayProvider.swift
Viscount88: No. It's bad example of protocol realization, because each provider confirm it and return action, but also this protocol confirm PurchaseAggregator - some providers wrapper, which return all actions from all providers. I didn't have time to fix it.

fearless/Modules/NewWallet/ChainAccount/ChainAccountPresenter.swift
Viscount88: I'll use providers aggregator here instead

fearless/Modules/Purchase/PurchaseViewFactory.swift
Viscount88: Wallet use old protocol method...

fearless/Modules/NewWallet/ChainAccount/ChainAccountPresenter.swift
Viscount88: I have no idea, to be honest. Context never used for no one implementation of that method

Viscount88: It always nil and seems like do nothing

bnsports (approved):

PR #483, opened 2022-02-01
(FLW-1515) Changing wallet name for wallet details screen

bnsports:


fearless/Modules/WalletDetails/WalletDetailsInteractor.swift
```diff
@@ -23,6 +26,34 @@ extension WalletDetailsInteractor: WalletDetailsInteractorInputProtocol {
     func setup() {
         fetchChainsWithAccounts()
     }
+
+    func update(_ walletName: String) {
```
bnsports, line 30: I think param name is required there, otherwise context may be missed, because interactor can update not only walletName

fearless/Modules/Profile/ProfilePresenter.swift
```diff
@@ -54,7 +54,10 @@ extension ProfilePresenter: ProfilePresenterProtocol {
         guard let wallet = selectedWallet else {
             return
         }
-        wireframe.showAccountDetails(from: view, metaAccount: wallet)
+        let walletNameChangeCompletion: (MetaAccountModel) -> Void = { [weak self] wallet in
```
bnsports, line 57: We have already begun a discussion that should be continued. I find the use of "*ModuleOutput" protocols much more convenient

fearless/Modules/WalletDetails/WalletDetailsInteractor.swift
Viscount88: Agree, I changed function name several times and missed it

fearless/Modules/Profile/ProfilePresenter.swift
Viscount88: This is modal screen, which can be opened from three different places. Completion seems the best option for me in that case.

Viscount88:
I had to use the event service instead of the output, because we need a global update point. The wallet name can be changed, for example, from a wallet tab, but we also need to know about that in profile module.

Viscount88 (approved):

PR #489, opened 2022-02-08
Hotfix/2.0.2

bnsports:
Fixed a bug that blocks user to go under existential deposit


fearless/Modules/NewWallet/WalletSendConfirm/WalletSendConfirmViewFactory.swift
```diff
@@ -53,6 +53,7 @@ struct WalletSendConfirmViewFactory {
             localizationManager: LocalizationManager.shared
         )
 
+        dataValidatingFactory.view = view
```
Viscount88, line 56: Would be better if factory won't know about view controller

Viscount88 (approved):

PR #490, opened 2022-02-09
[FLW-1461] Delete wallet check pin code

Viscount88:


fearless/Modules/Pincode/CheckPincode/CheckPincodePresenter.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    func didChangeState(from _: LocalAuthInteractor.LocalAuthState) {}
+
+    func didCompleteAuth() {
+        DispatchQueue.main.async { [weak self] in
+            self?.moduleOutput.didCheck()
+        }
+    }
+
+    func didUnexpectedFail() {}
```
bnsports, line 50: shouldn't we present error here?

Viscount88: We show error in didEnterWrongPincode() method. "didUnexpectedFail()" method was used only for case then user totally forgot his pin and want to sign up again.

bnsports (approved):

PR #538, opened 2022-03-21
Feature/flw 1361 about screen updates

DRadmir:


fearless/Modules/About/AboutPresenter.swift
```diff
@@ -60,14 +45,121 @@ extension AboutPresenter: AboutPresenterProtocol {
...
+extension AboutPresenter: Localizable {
+    func applyLocalization() {
+        view?.didReceive(locale: selectedLocale)
+        view?.didReceive(viewModel: aboutRows())
+    }
+}
+
+extension AboutPresenter {
+    // swiftlint:disable function_body_length
+    private func aboutRows() -> [AboutViewModel] {
```
bnsports, line 69: please create *ViewModelFactory class and move this code to it

fearless/Modules/About/AboutViewController.swift
```diff
@@ -1,236 +1,106 @@
...
-                return [.website, .opensource, .social, .writeUs, .terms, .privacy]
-            }
-        }
-    }
+    // MARK: - Private properties
 
-    var presenter: AboutPresenterProtocol!
+    private var locale = Locale.current
+    private var presenter: AboutPresenterProtocol
+    private var viewModel: [AboutViewModel]?
```
bnsports, line 18: instead of this we always use *ViewState pattern. so you need to create AboutViewState enum, please take a look "WalletSend" module for example

fearless/Modules/About/View/AboutTableViewCell.swift
```diff
@@ -0,0 +1,87 @@
+import UIKit
+import SnapKit
+
+final class AboutTableViewCell: UITableViewCell {
+    // MARK: - UI
+
+    private let titleLabel: UILabel! = {
```
bnsports, line 7: why force unwrap ? please remove it

bnsports (approved):

Viscount88 (approved):

PR #544, opened 2022-03-23
(flw 1610) two step import flow

Viscount88:


fearless/Modules/AccountCreate/AccountCreateViewController.swift
```diff
@@ -129,70 +85,55 @@ final class AccountCreateViewController: UIViewController {
...
+        if field == rootView.substrateDerivationPathField {
+            return substrateDerivationPathModel
+        } else if field == rootView.ethereumDerivationPathField {
+            return ethereumDerivationPathModel
         }
+        return nil
     }
 
-    @IBAction private func actionNext() {
+    @objc func actionNext() {
```
bnsports, line 126: why not private?

fearless/Modules/AccountCreate/AccountCreateViewLayout.swift
```diff
@@ -0,0 +1,343 @@
...
+
+private extension AccountCreateViewLayout {
+    private func configure() {
+        contentView.stackView.arrangedSubviews.forEach { $0.backgroundColor = R.color.colorBlack() }
+        advancedContainerView.isHidden = !expandableControl.isActivated
+        expandableControl.addTarget(self, action: #selector(actionExpand), for: .touchUpInside)
+    }
+
+    @objc
+    private func actionExpand() {
```
bnsports, line 223: I think it's better to use *ViewLayout class as layout and don't store any logic there. This piece of code should be replaced to ViewController class. agree?

bnsports (approved):

PR #550, opened 2022-03-25
FLW-1703: Added new viper template

DRadmir:


GenerambaTemplates/viper-code-layout/Code/presenter.swift.liquid
```diff
@@ -0,0 +1,38 @@
...
+        self.interactor = interactor
+        self.router = router
+        self.localizationManager = localizationManager
+    }
+    
+    // MARK: - Private methods
+}
+
+// MARK: - {{ module_info.name }}PresenterProtocol
+extension {{ module_info.name }}Presenter: {{ module_info.name }}PresenterProtocol {
```
Viscount88, line 25: Presenter protocol = ViewOutput + ModuleInput/Output protocols. I think we should explicitly announce protocol purpose.

GenerambaTemplates/viper-code-layout/Code/protocol.swift.liquid
```diff
@@ -0,0 +1,13 @@
+protocol {{ module_info.name }}ViewProtocol: AnyObject {}
+
+protocol {{ module_info.name }}PresenterProtocol: AnyObject {
```
Viscount88, line 3: It's ViewOutput protocol.

GenerambaTemplates/viper-code-layout/Code/viewlayout.swift.liquid
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+final class {{ module_info.name }}ViewLayout: UIView {
```
Viscount88, line 3: We have setupLayout() and configure() methods in each(or almost) ViewLayout class. So I think we can add it to private extension and remove if we for some reasons don't need them for current implementation

Viscount88 (approved):

PR #637, opened 2022-05-01
Feature/flw 1826 skeletons

DRadmir:


fearless/Modules/NewWallet/ChainAccountBalanceList/ChainAccountBalanceListViewLayout.swift
```diff
@@ -172,10 +192,62 @@ final class ChainAccountBalanceListViewLayout: UIView {
...
+    private func setupSkeleton() {
+        guard frame.size != .zero else { return }
+        let spaceSize = frame.size
+
+        let skeletonView = Skrull(
+            size: spaceSize,
+            decorations: [],
+            skeletons: createSkeletons(for: spaceSize)
+        )
+        .fillSkeletonStart(R.color.colorSkeletonStart()!)
```
Viscount88, line 230: Fill start and fill shouldn't be separated from each other. Also it possible that gradient contains more than two colors, so I think it'll be useful to provide array of colors, which can be converted to any gradient

Viscount88 (approved):

fearless/Modules/NewWallet/ChainAccountBalanceList/ChainAccountBalanceListViewLayout.swift
DRadmir: This methods implemented in pod SoraUI

PR #671, opened 2022-06-06
Fixed all parsing issues with assets and chains

posplaw:
Made set of changes in order to make app work fine even with broken JSONs like assets.json and chains.json

- Made `assets.json` and `chains.json` to be parsed as arrays without errors, skipping failing objects, and empty array if file is totally broken
- Prepared UX, so after PIN screen, user goes to main screen, but error should appear (**NOW NOT APPEARS**)
- If any of `ExternalApi` in ChainModel fails, continues parsing without that object
- Made `options` and `customNodes` to be always non-nil set/array, but with empty results, if nothing could be parsed (even with failing objects)
- Not existing options not parsed
- Added `Decodable` initializers in `ChainModel` and `ChainAssetModel` to specially unwrap potentially failable and not-required objects to don't fail deserialize of the Chain
- Added few extensions for decoding JSON to use in that case
- Added some more syntax sugar
- Added error when loaded chains without assets
- Skipping sync in case when chains loaded without assets


fearless/Common/Model/ChainRegistry/ChainModel.swift
```diff
@@ -35,12 +50,51 @@ class ChainModel: Codable {
...
 
+    enum DecodingError: Error {
+        case missingAssets
+        case missingNodes
+    }
+
+    required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        // Rules for app stable work
```
Viscount88, line 67: Why you placed this rules here?

fearless/Common/Model/ChainRegistry/ChainModel.swift
```diff
@@ -35,12 +50,51 @@ class ChainModel: Codable {
...
+
+        if assets.isEmpty {
+            throw DecodingError.missingAssets
+        }
+
+        if nodes.isEmpty {
+            throw DecodingError.missingNodes
+        }
+
+        // TODO: why types optional?
```
Viscount88, line 95: Already done?

Viscount88 (approved):

fearless/Common/Model/ChainRegistry/ChainModel.swift
posplaw: Yeah, types is optional because it's backward-compatibility for old runtime metadatas, when remote types file not used

fearless/Common/Configs/ApplicationConfigs.swift
```diff
@@ -158,16 +158,30 @@ extension ApplicationConfig: ApplicationConfigProtocol {
 
     // MARK: - GitHub
 
+    private var localFiles: Bool { true }
```
bnsports, line 161: useLocalFiles

bnsports: and false value in git I think

fearless/Common/Codable/JSON+Optional.swift
```diff
@@ -0,0 +1,26 @@
...
+    let base: T?
+
+    init(from decoder: Decoder) throws {
+        let decoder = try decoder.singleValueContainer()
+        base = try? decoder.decode(T.self)
+    }
+}
+
+extension KeyedDecodingContainer {
+    func decodeOptionalArray<T: Decodable>(_: [T].Type, forKey key: K) -> [T] {
```
bnsports, line 13: decodeOptionalArray from the name means that you have optional as result. decodeRequiredArray or something like decodeWithUnwrap should be there IMO

Viscount88 (approved):

Viscount88 (approved):

PR #735, opened 2022-08-05
Feature/flw 2091 network switch UI

DRadmir:


fearless/Common/View/SelectedNetworkButton.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import UIKit
+import SoraUI
+
+final class SelectedNetworkButton: UIButton {
+    var dotIsVisible: Bool = true {
```
bnsports, line 6: dot is always visible (comment from designer)

bnsports (approved):

PR #764, opened 2022-09-15
(FLW-2347) chain asset list view model building fix

Viscount88:


fearless/Modules/NewWallet/ChainAssetList/ViewModel/ChainAssetListViewModelFactory.swift
```diff
@@ -363,6 +348,16 @@ private extension ChainAssetListViewModelFactory {
 
         return priceWithChangeAttributed
     }
+
+    func filterUnique(chainAssets: [ChainAsset]) -> [ChainAsset] {
```
bnsports, line 352: filteredUnique. 'filter' sounds like void function

bnsports (approved):

PR #791, opened 2022-10-17
Feature/flw 2405 performance update to fix slow cold hot launch

DRadmir:


fearless/Common/Helpers/FWKeychainManager.swift
```diff
@@ -0,0 +1,118 @@
+import Foundation
+import SoraKeystore
+
+final class FWKeychainManager {
```
bnsports, line 4: please modify this class in common repository. just pass queue to constructor e.g.

fearless/Common/Substrate/Types/CallCodingPath.swift
```diff
@@ -1,52 +1,75 @@
...
+            return (moduleName: "NominationPools", callName: "set_metadata")
+        case .poolBondMore:
+            return (moduleName: "NominationPools", callName: "bond_extra")
+        case .poolUnbond:
+            return (moduleName: "NominationPools", callName: "unbond")
+        case .claimPendingRewards:
+            return (moduleName: "NominationPools", callName: "claim_payout")
+        case .poolWithdrawUnbonded:
+            return (moduleName: "NominationPools", callName: "withdraw_unbonded")
+        case let .fromInit(moduleName, callName):
```
bnsports, line 54: as I understand we need this case because there are places in our code that not cover by known enum cases (like .transfer or .addMemo). so I suggest to find those cases and cover with enum cases. it's important because that way we can avoid manual allCases usage.

fearless/Modules/Pincode/CheckPincode/CheckPincodePresenter.swift
```diff
@@ -43,7 +43,7 @@ extension CheckPincodePresenter: LocalAuthInteractorOutputProtocol {
         }
     }
 
-    func didChangeState(from _: LocalAuthInteractor.LocalAuthState) {}
+    func didChangeState(to _: LocalAuthInteractor.LocalAuthState) {}
```
bnsports, line 46: unused parameter

fearless/Modules/Pincode/LocalAuthentification/LocalAuthPresenter.swift
```diff
@@ -55,11 +55,17 @@ class LocalAuthPresenter: PinSetupPresenterProtocol {
 extension LocalAuthPresenter: LocalAuthInteractorOutputProtocol {
     func didEnterWrongPincode() {
         DispatchQueue.main.async { [weak self] in
+            self?.view?.didStopLoading()
```
bnsports, line 58: I didn't see 'didStopLoading' for any other case except wrong. please check and add if needed

fearless/Modules/Pincode/ScreenAuthorization/ScreenAuthorizationPresenter.swift
```diff
@@ -45,7 +45,7 @@ extension ScreenAuthorizationPresenter: LocalAuthInteractorOutputProtocol {
         }
     }
 
-    func didChangeState(from _: LocalAuthInteractor.LocalAuthState) {}
+    func didChangeState(to _: LocalAuthInteractor.LocalAuthState) {}
```
bnsports, line 48: unused parameter

fearless/Common/Substrate/UsedRuntimeModules.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+struct UsedRuntimePaths {
+    private let storageCodingCases = StorageCodingPath.allCases
+    private let callCodingCases = CallCodingPath.allCases
+    private let constantsCodingCases = ConstantCodingPath.allCases
+    private let sabstrateCodingCases = SubstrateCallPath.allCases
+
+    lazy var usedRuntimePaths: [String: [String]] = {
```
bnsports, line 9: looks not optimzed. can you think about use compactMap and reduce along with Set, instead of 'forEach'. It should works faster and looks better

bnsports (changes requested):

fearless/Modules/Pincode/CheckPincode/CheckPincodePresenter.swift
DRadmir: yes, this is LocalAuthInteractorOutputProtocol

fearless/Modules/Pincode/LocalAuthentification/LocalAuthPresenter.swift
DRadmir: this is not needed, because if pin valid we join in the main screen

fearless/Modules/Pincode/ScreenAuthorization/ScreenAuthorizationPresenter.swift
DRadmir: yes, this is LocalAuthInteractorOutputProtocol

fearless/Common/Substrate/UsedRuntimeModules.swift
DRadmir: forEach is faster than reduce, compactMap is not beautiful for this case
https://www.skoumal.com/en/performance-of-built-in-higher-order-functions-map-filter-reduce-and-flatmap-vs-for-in-loop-in-swift/

bnsports (approved):

PR #800, opened 2022-10-27
(flw 2471) keyboard adoptable

Viscount88:


fearless/Common/Protocols/KeyboardAdoptable.swift
```diff
@@ -14,21 +14,16 @@ extension KeyboardAdoptable {
...
         keyboardHandler = nil
     }
 }
 
-protocol KeyboardViewAdoptable: KeyboardAdoptable {
-    var target: UIView? { get }
-    var currentKeyboardFrame: CGRect? { get set }
-    var shouldApplyKeyboardFrame: Bool { get }
+protocol KeyboardViewAdoptable: KeyboardAdoptable, KeyboardHandlerDelegate {
+    var target: Constraint? { get }
```
DRadmir, line 26: what if I use frame for view? I can use offsetFromKeyboardWithInset?

DRadmir: And we can't open pull request in SoraFoundation, because independent from SnapKit

DRadmir (changes requested):

fearless/Common/Helpers/FearlessKeyboardHandler.swift
```diff
@@ -0,0 +1,97 @@
...
+    var handleWillShow: KeyboardFrameChangeHandleBlock?
+    var handleWillHide: KeyboardFrameChangeHandleBlock?
+
+    // MARK: Initialization
+
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
+    public convenience init() { self.init(with: nil) }
```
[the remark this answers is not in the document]
Viscount88: we have no delegate, but handlers for keyboard still working

fearless/Common/Protocols/KeyboardAdoptable.swift
Viscount88: KeyboardAdoptable is not part of SoraFoundation, only KeyboardHandler. Use constraint is more safety and obvious than uiview

PR #807, opened 2022-10-31
Fix/flw 2460 nodes not switching via network issues

DRadmir:


fearless/Modules/NodeSelection/NodeSelectionInteractor.swift
```diff
@@ -20,12 +20,27 @@ final class NodeSelectionInteractor {
...
+
+        operationManager.enqueue(operations: [operation], in: .transient)
+    }
 }
 
 extension NodeSelectionInteractor: NodeSelectionInteractorInputProtocol {
     func setup() {
-        presenter?.didReceive(chain: chain)
-
+        fetchChainModel()
```
bnsports, line 43: why?

bnsports (approved):

PR #850, opened 2022-11-24
(FLW-2469) QR scan encoder/decoder for sora and others

Viscount88:


fearless/Common/QRCoding/QRCreationOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+                result = .failure(QRCreationOperationError.generatedImageInvalid)
+            }
+
+            return
+        }
+
+        let transformedImage: CIImage
+
+        if qrImage.extent.size.width * qrImage.extent.height > 0.0 {
+            let transform = CGAffineTransform(
```
bnsports, line 47: didn't you miss original image ratio there?
I think you need to calculate one scale for both x and y axis

fearless/Common/QRCoding/QRService.swift
```diff
@@ -0,0 +1,50 @@
...
+
+final class QRService {
+    let operationFactory: QROperationFactoryProtocol
+    let operationQueue: OperationQueue
+
+    private let encoder: QREncoderProtocol
+
+    public init(
+        operationFactory: QROperationFactoryProtocol,
+        encoder: QREncoderProtocol = QREncoder(),
```
bnsports, line 21: I think it will be more beautiful in case you pass concrete encoder implementation there. This constructor takes protocol as parameter, why not ?

fearless/Common/QRCoding/QRService.swift
```diff
@@ -0,0 +1,50 @@
...
+        self.operationFactory = operationFactory
+        self.encoder = encoder
+        self.operationQueue = operationQueue
+    }
+}
+
+extension QRService: QRServiceProtocol {
+    @discardableResult
+    func generate(
+        with qrType: QRType,
```
bnsports, line 33: you can avoid passing this parameter in case of passing concrete encoder implementation in this class.

fearless/Modules/Wallet/ReceiveAsset/ReceiveAssetPresenter.swift
```diff
@@ -105,20 +104,24 @@ private extension ReceiveAssetPresenter {
...
         }
-
-        let receiveInfo = ReceiveInfo(
-            accountId: accountId.toHex(),
-            assetId: asset.id,
-            amount: nil,
-            details: nil
-        )
+        var qrType: QRType = .address(address)
+        if chain.isSora {
```
bnsports, line 112: need to think what to do there in case of my suggestions will be implemented

bnsports (approved):

fearless/Common/QRCoding/QRCreationOperation.swift
Viscount88: It's okay. This logic is already checked.

PR #882, opened 2023-01-20
Feature/flw 2680 sora sq project

DRadmir:


fearless/Modules/NewWallet/WalletTransactionHistory/WalletTransactionHistoryViewFactory.swift
```diff
@@ -93,4 +84,37 @@ enum WalletTransactionHistoryViewFactory {
             items: filters
         )]
     }
+
+    private static func createHistoryDeps(
```
bnsports, line 88: actually historyoperationfactory used across the app, not only inside this module. because of that, it's better to have some Assembly that can generate concrete implementation of history operation factory based on chain. I'm already implemented this assembly so it will be replaced once we merged our branches together. nothing to do here right now.

fearless/en.lproj/Localizable.strings
```diff
@@ -970,3 +971,4 @@ Remember to make a backup of your key and keep it in a safe and private place (e
 "polkaswap.dex.alert.title" = "Token Pair Isn’t Created";
 "polkaswap.quotes.not.available" = "Quotes not available";
 "all.done.subscan.button.title" = "Subscan";
+"transaction.details.swap.awesome.wallet" = "My Awesome Wallet";
```
bnsports, line 974: whats is this

bnsports (changes requested):

fearless/Modules/NewWallet/WalletTransactionDetails/WalletTransactionDetailsViewLayout.swift
```diff
@@ -82,6 +82,8 @@ final class WalletTransactionDetailsViewLayout: UIView {
             if let extrinsicViewModel = viewModel as? ExtrinsicTransactionDetailsViewModel {
                 bindExtrinsic(viewModel: extrinsicViewModel)
             }
+        case .swap, .unused:
+            break
```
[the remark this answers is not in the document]
DRadmir: Because for swap we show new screen

bnsports (approved):

PR #889, opened 2023-01-26
FLW-2926: GLMR staking stake less fix. Any fixes

DRadmir:


fearless/Common/ViewModel/Amount/AmountInputViewModel.swift
```diff
@@ -27,7 +27,7 @@ final class AmountInputViewModel: AmountInputViewModelProtocol, MoneyPresentable
     }
 
     var decimalAmount: Decimal? {
-        if amount.isEmpty {
+        if amount.isEmpty || amount.last == "," || amount.last == "." {
```
Viscount88, line 30: mb add "0" instead of return nil?

Viscount88 (approved):

PR #905, opened 2023-02-03
(FLW-2880) send address validation

Viscount88:
handle same wallet address, select chain opportunity then address is invalid


fearless/Modules/Send/SendPresenter.swift
```diff
@@ -90,7 +90,9 @@ extension SendPresenter: SendViewOutput {
                 isValid: true
             )
             view.didReceive(viewModel: viewModel)
-            interactor.getPossibleChains(for: address)
+            interactor.getPossibleChains(for: address) { possibleChains in
+                didReceive(possibleChains: possibleChains)
```
DRadmir, line 94: please check this line, you don't use self or weak self

fearless/Modules/Send/SendPresenter.swift
```diff
@@ -113,14 +115,66 @@ extension SendPresenter: SendViewOutput {
...
     }
 
-    func didTapContinueButton() {
-        guard let chainAsset = selectedChainAsset else { return }
-        guard let address = recipientAddress,
-              interactor.validate(address: address, for: chainAsset.chain)
-        else {
-            router.present(message: nil, title: "Incorrect address", closeAction: "Close", from: view)
-            return
+    func validateAddress(with chainAsset: ChainAsset, successCompletion: @escaping (String) -> Void) {
```
DRadmir, line 118: This construction very cumbersome. Please separate this on the small private methods

DRadmir (changes requested):

DRadmir (approved):

PR #915, opened 2023-02-10
(flw 3025) color shadows for asset icons

Viscount88:


fearless/Common/Model/ChainRegistry/AssetModel.swift
```diff
@@ -64,6 +67,7 @@ struct AssetModel: Equatable, Codable, Hashable {
         currencyId = try? container.decode(String?.self, forKey: .currencyId)
         displayName = try? container.decode(String?.self, forKey: .displayName)
         existentialDeposit = try? container.decode(String?.self, forKey: .existentialDeposit)
+        color = try container.decode(String.self, forKey: .color)
```
bnsports, line 70: there should be optional 'try?' instead

bnsports (changes requested):

bnsports (approved):

PR #1168, opened 2023-11-28
nft collection module

Viscount88:


fearless/Modules/NFT/NftCollection/ViewModel/NftCollectionViewModelFactory.swift
```diff
@@ -1,24 +1,78 @@
...
-        return NftCollectionViewModel(collectionName: collection.displayName, cellModels: cellModels)
+        let ownedCellModels = buildCellModels(
+            from: collection.nfts ?? [],
+            type: .owned,
+            locale: locale
+        )
+
+        let availableCellModels = buildCellModels(
+            from: collection.availableNfts?.filter { nft in
+                collection.nfts?.contains(nft) != true
```
bnsports, line 35: it's better to compare with id

fearless/Modules/NFT/NftCollection/NftCollectionViewLayout.swift
```diff
@@ -60,13 +72,28 @@ final class NftCollectionViewLayout: UIView {
...
             make.leading.trailing.bottom.equalToSuperview()
-            make.top.equalTo(navigationBar.snp.bottom)
+            make.top.equalTo(titleLabel.snp.bottom)
         }
     }
 
     func bind(viewModel: NftCollectionViewModel) {
         navigationTitleLabel.text = viewModel.collectionName
+        viewModel.collectionImage?.loadImage(on: imageView, targetSize: CGSize(
+            width: UIScreen.main.bounds.width - 16,
```
bnsports, line 94: it's incorrect to use magic number here because it depends on leading and trailing insets of imageView. so please use local Constants enum

fearless/Modules/NFT/NftCollection/NftCollectionViewController.swift
```diff
@@ -72,26 +77,80 @@ extension NftCollectionViewController: UICollectionViewDataSource, UICollectionV
...
         let size: CGFloat = (rootView.collectionView.frame.size.width - space) / 2.0
-        return CGSize(width: size, height: 233)
+        return CGSize(width: size, height: 249)
     }
 
-    func collectionView(_: UICollectionView, numberOfItemsInSection _: Int) -> Int {
-        cellModels?.count ?? 0
+    func collectionView(_ collectionView: UICollectionView, viewForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> UICollectionReusableView {
+        if kind == UICollectionView.elementKindSectionHeader {
+            let sectionHeader = collectionView.dequeueReusableSupplementaryView(ofKind: kind, withReuseIdentifier: "Collection view section header", for: indexPath) as! CollectionViewSectionHeader
```
bnsports, line 85: reuse identifier should be more unique

fearless/Modules/NFT/NftCollection/NftCollectionRouter.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 final class NftCollectionRouter: NftCollectionRouterInput {
-    func openNftDetails(nft: NFT, wallet: MetaAccountModel, address: String, from view: ControllerBackedProtocol?) {
+    func openNftDetails(nft: NFT, type _: NftType, wallet: MetaAccountModel, address: String, from view: ControllerBackedProtocol?) {
```
bnsports, line 4: unused parameter type

fearless/CoreLayer/OperationFactory/NFT/AlchemyNFTOperationFactory.swift
```diff
@@ -20,7 +20,10 @@ final class AlchemyNFTOperationFactory {
         let authorizedUrl = url.appendingPathComponent(ThirdPartyServicesApiKeysDebug.alchemyApiKey)
         let endpointUrl = authorizedUrl.appendingPathComponent("getContractsForOwner")
         var urlComponents = URLComponents(string: endpointUrl.absoluteString)
-        urlComponents?.queryItems = [URLQueryItem(name: "owner", value: address)]
+        urlComponents?.queryItems = [
+            URLQueryItem(name: "owner", value: address),
+            URLQueryItem(name: "withMetadata", value: "true")
```
bnsports, line 25: what if there's no metadata? maybe we can extract at least nft name ?

bnsports (changes requested):

fearless/pt-PT.lproj/Localizable.strings
```diff
@@ -1228,3 +1228,5 @@ Mantenha, faça STAKE ou forneça liquidez de um valor mínimo de XOR equivalent
 "verify.phone.number.description" = "Introduzir o código SMS que foi enviado para\n%@";
 "verify.phone.number.send.code" = "ENVIAR CÓDIGO";
 "verify.phone.number.title" = "Verificar o seu número de telefone";
+"nft.collection.my.nfts" = "My NFTs";
+"nft.collection.available.nfts" = "Available NFTs in the %s collection";
```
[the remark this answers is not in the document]
Viscount88: Not localised yet

fearless/Modules/NFT/NftCollection/NftCollectionRouter.swift
Viscount88: It'll be used later

fearless/CoreLayer/OperationFactory/NFT/AlchemyNFTOperationFactory.swift
Viscount88: It doesn't work like a filter. Here we inform whether we want to load metadata for NFT or not. If not, the download is faster, but we need the metadata. The default value here was already true, I just indicated it more explicitly, if in the future we want to change something it will be obvious where to do it.

bnsports (approved):

PR #1177, opened 2023-12-22
Fixes/bug fixes

DRadmir:


fearless/Modules/Filters/FiltersViewLayout.swift
```diff
@@ -39,12 +41,17 @@ final class FiltersViewLayout: UIView {
 
     func applyLocalization() {
         resetButton.setTitle(R.string.localizable.commonReset(preferredLanguages: locale.rLanguages), for: .normal)
-        navigationTitleLabel.text = R.string.localizable.walletFiltersTitle(preferredLanguages: locale.rLanguages)
+        if screenTitle == nil {
```
Viscount88, line 44: We can use something like text = screenTitle ? R.string...

Viscount88 (changes requested):

Viscount88 (approved):

PR #1210, opened 2024-02-15
Release/3.3.4

bnsports:


fearless/Modules/Send/SendDependencyContainer.swift
```diff
@@ -144,7 +151,8 @@ final class SendDepencyContainer {
                 secretKeyData: secretKey,
                 cryptoType: SFCryptoType(utilsType: accountResponse.cryptoType.utilsType, isEthereum: chainAsset.chain.isEthereumBased)
             )
-            let callFactory = SubstrateCallFactoryAssembly.createCallFactory(forSSF: runtimeService.runtimeSpecVersion)
+
+            let callFactory = SubstrateCallFactoryDefault(runtimeService: nativeRuntimeService)
```
bnsports, line 155: looks like we should move those changes to ssf

fearless/Modules/Send/SendPresenter.swift
```diff
@@ -925,8 +935,10 @@ extension SendPresenter: SendInteractorOutput {
         switch result {
         case let .success(minimumBalance):
             self.minimumBalance = minimumBalance
+            view?.switchEnableSendAllVisibility(isVisible: true)
```
bnsports, line 938: we should also ensure that minimalBalance > 0

fearless/Modules/Send/SendProtocols.swift
```diff
@@ -18,6 +18,8 @@ protocol SendViewInput: ControllerBackedProtocol, LoadableViewProtocol {
     func didBlockUserInteractive(isUserInteractiveAmount: Bool)
     func setInputAccessoryView(visible: Bool)
     func setHistoryButton(isVisible: Bool)
+    func enableSendAll()
```
bnsports, line 21: let's think about sendAll property like about viewModel. it seems more accurate and beautiful (IMO):
func didReceive(sendAllEnabled: Bool)

fearless/Modules/Send/Validators/SendDataValidatingFactory.swift
```diff
@@ -104,20 +106,22 @@ class SendDataValidatingFactory: NSObject {
...
+                    warningHandler?()
                 },
                 locale: locale
             )
-
         }, preservesCondition: {
             guard !chainAsset.chain.isEthereum else {
                 return true
             }
+            if sendAllEnabled, canProceedIfViolated {
```
bnsports, line 122: looks wrong. should be sendAllEnabled || canProceedIfViolated

fearless/Modules/Send/Validators/SendDataValidatingFactory.swift
```diff
@@ -204,7 +211,8 @@ class SendDataValidatingFactory: NSObject {
                 return
             }
             let title = R.string.localizable.commonWarning(preferredLanguages: locale.rLanguages)
-            let message = R.string.localizable.soraBridgeAmountLessFee(preferredLanguages: locale.rLanguages)
+            let originKnownChain = Chain(chainId: originCHainId)?.rawValue ?? ""
```
bnsports, line 214: there's better to have some default value instead of empty string. "Uknown" maybe?

fearless/Modules/WalletDetails/WalletDetailsPresenter.swift
```diff
@@ -116,6 +116,8 @@ extension WalletDetailsPresenter: WalletDetailsInteractorOutputProtocol {
             else { return }
             let action = items[selectedIndex]
             switch action {
+            case .claimCrowdloanRewards:
+                break
```
bnsports, line 120: is action needed?

fearless/Modules/Send/Validators/SendDataValidatingFactory.swift
Viscount88: no, if we enable full send AND chain can proceed violation - then we can do it, if anything isn't available - we can't

fearless/Modules/Send/SendProtocols.swift
Viscount88: I'm not sure, it's just reaction when user set enable from alert, it's always true

Viscount88 (approved):

PR #1252, opened 2024-06-04
Feature/flw 4635 new bridge route

DRadmir:


fearless/Modules/Send/Validators/SendDataValidatingFactory.swift
```diff
@@ -299,4 +317,29 @@ class SendDataValidatingFactory: NSObject {
...
     }
+
+    private func minAssetAmount(
+        originCHainId: ChainModel.Id,
+        destChainId: ChainModel.Id
+    ) -> String {
+        let originKnownChain = Chain(chainId: originCHainId)
+        let destKnownChain = Chain(chainId: destChainId)
+
+        switch (originKnownChain, destKnownChain) {
```
bnsports, line 328: what if some of those chains (polkadot, kusama, acala, liberland) will add 'XOR' asset and you call this method to determine minimum amount of xor to transfer from polkadot to sora ?

bnsports (approved):

